PR #27, opened 2015-02-06
Refactor ASTKernel into separate classes

chromy:
Are we okay to land this now?
The corresponding PyOP2 change is here: https://github.com/OP2/PyOP2/pull/422


miklos1:
Closing, since this seems out of date. If it's still relevant, please propose the changes again.

PR #35, opened 2015-03-04
Inverse kernel

dham:
Provide a functioning invert node in the Coffee tree. This is required for both inverse assembly and FInAT inverse Jacobians.


FabioLuporini:
We are good to go. Merging.

PR #36, opened 2015-03-06
Padding allocates suitable buffers

FabioLuporini:
If padding is requested, input data is suitably mapped to a padded, temporary array, allowing the caller to completely abstract from this optimization

### DIFF
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -906,7 +906,10 @@ class Access(object):
     _modes = ["READ", "WRITE", "RW", "INC", "DEC", "IMUL", "IDIV"]
 
     def __init__(self, mode):
-        self.mode = mode
+        self._mode = mode
+
+    def __eq__(self, other):
+        return self._mode == other._mode
 
 
 READ = Access("READ")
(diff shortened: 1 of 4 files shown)


wence-:
I'd like to land OP2/PyOP2#430, is this side of the change good to go?

coffee/base.py
```diff
@@ -833,3 +849,26 @@ def c_flat_for(code, parent):
...
+    def __init__(self, mode):
+        self._mode = mode
+
+    def __eq__(self, other):
+        return self._mode == other._mode
+
+
+READ = Access("READ")
+WRITE = Access("WRITE")
+Rw = Access("RW")
```
kynan: Why not RW?

FabioLuporini: I just misspelled it, but there's already a commit fixing it on another branch (ast-if-nodes), which I hope will land soon

FabioLuporini:
@wence-  Yes, this is good to go, feel free to merge it as soon as you feel ready to merge the pyop2 changes (yes, they must land together)

wence-:
I've fixed this to DTRT in the presence of Invert nodes and rebased on master.

PR #37, opened 2015-03-23
Determinant kernels

mlange05:
Adds special-case implementation nodes for determinant computation of 1x1, 2x2 and 3x3 square matrices, as well as a node for absolute values.

### DIFF
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -512,9 +512,9 @@ class Invert(Statement, Perfect, Linalg):
   int info;
 
   dgetrf_(&n,&n,%s,&lda,ipiv,&info);
-  dgetri_(&n,A,&lda,ipiv,work,&lwork,&info);
+  dgetri_(&n,%s,&lda,ipiv,work,&lwork,&info);
 }
-""" % (str(dim), str(lda), str(sym))
+""" % (str(dim), str(lda), str(sym), str(sym))
 
 
 class Decl(Statement, Perfect):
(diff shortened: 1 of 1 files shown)


FabioLuporini:
Hey Micheal, here are a few comments:
- can we give the Det classes more meaningful names?
- can we not use FunCall for generic function calls, instead of specific classes, like Fabs?
Thanks!

mlange05:
Hi Fabio, 
Concerning Fabs I would argue that the `Abs` node should be part of the API because Coffee, which  is aware of the target architecture, is in a much better position to decide on implementation details than user applications like FInAT. However, I am happy to pull this out for now since either way works for the current problem.

For the Det class names, do you have any suggestions? Thanks

FabioLuporini:
I was thinking that in 99.9% of the cases any time you have to evaluate {f}abs you reduce to call a standard math.h function. If that is the case, I would prefer to stick to generic FunCall. On the other hand, if you have more specific examples in mind, I'm open to discussion :)

As for the Dets, mmm...how bad do Determinant1x1, Determinant2x2 and Determinant3x3 sound? I know most of the class names are abbreviated (like Sub or Div etc.) but for LinAlg objects, which are in a certain sense special, I would prefer to keep them expanded. Also, from the moment we are soon landing this, what do you think about gathering all LinAlg classes at the same point of the file? So this would include Det classes and Invert, at the moment, but we will have more in the future. These, of course, are just minor comments

mlange05:
Ok, sounds good. I'll do the rebase as per you suggestions. Thanks

FabioLuporini:
Thank you!

mlange05:
Done

FabioLuporini:
OK, this is good to go, merging. Thanks.

PR #39, opened 2015-03-31
Additional features and enhanced support for kernel optimisation

FabioLuporini:
This PR:
- Fixes a few bugs related to AST transformation
- Extends the AST hierarchy
- Changes the way padding is performed to handle the case in which offsets are present
- Fixes the expression rewriter when multiple expressions are found in the kernel
- Generally improves the code quality

### DIFF
--- a/coffee/utils.py
+++ b/coffee/utils.py
@@ -403,6 +403,8 @@ def visit(node, parent=None, search=None):
             info['symbols_dep'][node] = dep
             info['symbols_mode'][node] = access_mode
             info['symbol_refs'][node.symbol].append((node, parent))
+            for r in node.rank + node.offset:
+                inspect(r, node)
         elif isinstance(node, Expr):
             for child in node.children:
                 inspect(child, node)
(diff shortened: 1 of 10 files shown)


wence-:
Does this subsume #36, or does that need to go first?

FabioLuporini:
#36 and the corresponding changes in PyOP2 should go first

coffee/utils.py
```diff
@@ -100,6 +100,70 @@ def unroll_factors(loops):
...
+        def mark(node):
+            if Process.start is not None:
+                Process._processed.append((node, Process.start, Process.end,
+                                           Process.decls, Process.blockable))
+            Process.start = None
+            Process.end = None
+            Process.decls = {}
+            Process.blockable = []
+
+    def init_decl(stmt):
```
kynan: Why not make those methods of `Process`?

FabioLuporini:
A tiny note before I forget to say: @wence- , this PR (I've already rebased it) also works around the disgusting hack you had to do for padding in presence of inverse

PR #41, opened 2015-04-21
Lower layers loop

doru1004:
Add more AST node types.

### DIFF
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -108,6 +108,10 @@ class Node(object):
     def pragma(self):
         return self._pragma
 
+    @property
+    def has_children(self):
+        return bool(self.children)
+
     @pragma.setter
     def pragma(self, _pragma):
         self._pragma = self._format_pragma(_pragma)
@@ -350,6 +354,10 @@ class Cast(UnaryExpr):
     def __init__(self, cast_type, expr):
         super(Cast, self).__init__(expr)
         self.cast_type = as_symbol(cast_type)
+
+    def __deepcopy__(self, memo):
+        """This copy method allows the passing in of a cast type."""
+        return self.__class__(self.cast_type, dcopy(self.children[0]))
 
     def gencode(self, scope=True):
         return "(%s) %s" % (str(self.cast_type), wrap(self.children[0].gencode()))
@@ -835,29 +843,6 @@ class For(Statement):
         elif isinstance(self.init, Assign):
             return self.init.children[0]
 
-    def set_itvar(self, new_var):
-        # Change the name of the index init
-        if isinstance(self.init, Decl):
-            self.init.sym.symbol = new_var
-        elif isinstance(self.init, Assign):
-            self.init.children[0].symbol = new_var
-        else:
-            raise NotImplemented("Don't know how to change index name in this init node.")
-        # Change the name of the index in the comparison
-        if isinstance(self.cond.children[0], Symbol):
-            self.cond.children[0].symbol = new_var
-        elif isinstance(self.cond.children[0], str):
-            self.cond.children[0] = new_var
-        else:
-            raise NotImplemented("Don't know how to change index name in this condition node.")
-        # Change the name of the index in the increment
-        if isinstance(self.incr.children[0], Symbol):
-            self.incr.children[0].symbol = new_var
-        elif isinstance(self.incr.children[0], str):
-            self.incr.children[0] = new_var
-        else:
-            raise NotImplemented("Don't know how to change index name in this increment node.")
-
     @property
     def start(self):
         return self.init.rvalue.symbol
@@ -936,7 +921,6 @@ class If(Statement):
     """
 
     def __init__(self, if_expr, branches):
-        # TODO: Re-implement this using two arguments a then branch and an optional else branch
         self.then_body = branches[0]
         if not isinstance(branches[0], Node):
             self.then_body = Block(branches[0], open_scope=True)
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -343,6 +343,18 @@ class Deref(UnaryExpr):
         return "*%s" % wrap(self.children[0].gencode())
 
 
+class Cast(UnaryExpr):
+
+    """Cast a variable to a different type."""
+
+    def __init__(self, cast_type, expr):
+        super(Cast, self).__init__(expr)
+        self.cast_type = as_symbol(cast_type)
+
+    def gencode(self, scope=True):
+        return "(%s) %s" % (str(self.cast_type), wrap(self.children[0].gencode()))
+
+
 class Addr(UnaryExpr):
 
     """Get the address."""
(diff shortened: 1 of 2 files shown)


coffee/base.py
```diff
@@ -663,6 +731,24 @@ def gencode(self, not_scope=False):
...
+        # If the then branch is a plain list, cast it to a Block.
+        if not isinstance(then_body, Node):
+            then_body = Block(then_body, open_scope=True)
+        super(Ifthen, self).__init__([then_body])
+        self.cond = cond
+
+    def gencode(self, scope=True):
+        return "if (" + str(self.cond) + ")\n" + indent(self.children[0].gencode())
+
+
```
wence-: OK, I see you renamed to Ifthen, from If.  But I'm confused why you need this?

coffee/base.py
```diff
@@ -216,6 +216,30 @@ def gencode(self, not_scope=True):
...
+class Addr(UnaryExpr):
+
+    """Get the address."""
+
+    def gencode(self):
+        return "&%s" % wrap(self.children[0].gencode())
+
+
```
dorugeber: COFFEE is just turning into verbose-C?

coffee/base.py
```diff
@@ -216,6 +216,30 @@ def gencode(self, not_scope=True):
...
+    def gencode(self):
+        return "*%s" % wrap(self.children[0].gencode())
+
+
+class Cast(UnaryExpr):
+
+    """Dereference an address."""
+
+    def gencode(self):
+        return "(%s) %s" % wrap(self.children[0].gencode())
```
wence-: Does wrap return two strings?  I can't understand how this works?

coffee/base.py
doru1004: Initially there was no IF and I added an If with only the then branch present. In the meantime another If was added with both then and else branches. Do you want me to code this IfThen into the If node?

coffee/base.py
doru1004: Well we have to support all the unary operators supported by C.

coffee/utils.py
```diff
@@ -643,3 +644,72 @@ def set_itspace(loop_a, loop_b):
...
+        if var_name == old_name:
+            kernel_ast = scopy(kernel_ast)
+            kernel_ast.set_itvar(new_name)
+    children = []
+    for child in kernel_ast.children:
+        children.append(rename_var(child, old_name, new_name))
+    kernel_ast.children = children
+    return kernel_ast
+
+def contains_flat_block(tree):
```
FabioLuporini: You don't need this, just use visit(node, search=FlatBlock)

coffee/utils.py
```diff
@@ -643,3 +644,72 @@ def set_itspace(loop_a, loop_b):
...
+    # Check if the AST contains any FlatBlock nodes
+    if isinstance(tree, FlatBlock):
+        return True
+    elif not isinstance(tree, str):
+        for child in tree.children:
+            if contains_flat_block(child):
+                return True
+    return False
+
+def change_kernel_loop_index_ids(tree):
```
FabioLuporini: I think there's already another utility function for doing this

coffee/utils.py
```diff
@@ -643,3 +644,72 @@ def set_itspace(loop_a, loop_b):
 def loops_as_dict(loops):
     loops_itvars = [l.itvar for l in loops]
     return OrderedDict(zip(loops_itvars, loops))
+
+def rename_var(kernel_ast, old_name, new_name):
```
FabioLuporini: Can't you just use the utility function ast_replace?

coffee/utils.py
```diff
@@ -643,3 +644,72 @@ def set_itspace(loop_a, loop_b):
...
+    This method renames the variables represented by a Symbol object.
+    The Symbol's string is replaces the ``old_name`` with the
+    new string ``new_name``.
+    """
+    # Warning: Inconsistencies in the representation could make this
+    # function fail if variable names are polluted with
+    # *, & and [] symbols.
+    if isinstance(kernel_ast, Symbol):
+        var_name = str(kernel_ast.symbol)
+        if "*" in var_name or "[" in var_name or "]" in var_name or "&" in var_name:
```
FabioLuporini: We have to find a better way of doing this

coffee/utils.py
```diff
@@ -643,3 +644,72 @@ def set_itspace(loop_a, loop_b):
...
+        return kernel_ast
+    if not kernel_ast.has_children:
+        return kernel_ast
+    if isinstance(kernel_ast, For):
+        var_name = str(kernel_ast.itvar)
+        if "*" in var_name or "[" in var_name or "]" in var_name or "&" in var_name:
+            raise RuntimeError("The for loop index name name has been polluted with * & [ or ] characters.")
+        if var_name == old_name:
+            kernel_ast = scopy(kernel_ast)
+            kernel_ast.set_itvar(new_name)
```
FabioLuporini: Also, looks like copy-paste

coffee/base.py
```diff
@@ -683,22 +749,30 @@ def gencode(self):
...
+    """If-then-else construct.
 
     :param if_expr: The expression driving the jump
-    :param branches: A 2-tuple of AST nodes, respectively the 'if' and the 'else'
+    :param branches: A 2-tuple of AST nodes, respectively the 'then' and the 'else'
                      branches
     """
 
     def __init__(self, if_expr, branches):
+        # TODO: Re-implement this using two arguments a then branch and an optional else branch
```
FabioLuporini: I was avoiding TODOs in master... What are you trying to do here?

coffee/base.py
```diff
@@ -683,22 +749,30 @@ def gencode(self):
...
+        self.then_body = branches[0]
+        if not isinstance(branches[0], Node):
```
FabioLuporini: Why shouldn't they be Nodes?

coffee/base.py
```diff
@@ -683,22 +749,30 @@ def gencode(self):
...
+            self.then_body = Block(branches[0], open_scope=True)
+        if len(branches) == 2:
+            self.else_body = branches[1]
+            if not isinstance(branches[1], Node):
```
FabioLuporini: Why shouldn't they be Nodes?

coffee/base.py
```diff
@@ -632,6 +672,32 @@ def itvar(self):
...
+        if isinstance(self.cond.children[0], Symbol):
+            self.cond = scopy(self.cond)
+            self.cond.children[0].symbol = new_var
+        elif isinstance(self.cond.children[0], str):
+            self.cond.children[0] = new_var
+        else:
+            raise NotImplemented("Don't know how to change index name in this condition node.")
+        # Change the name of the index in the increment
+        if isinstance(self.incr.children[0], Symbol):
+            self.incr = scopy(self.incr)
```
FabioLuporini: shallow copies are dangerous. Why are you using them?

coffee/base.py
```diff
@@ -632,6 +672,32 @@ def itvar(self):
...
+    def set_itvar(self, new_var):
+        # Change the name of the index init
+        self.init = scopy(self.init)
+        if isinstance(self.init, Decl):
+            self.init.sym.symbol = new_var
+        elif isinstance(self.init, Assign):
+            self.init = Decl("int", Symbol(new_var), self.init.children[1])
+        else:
+            raise NotImplemented("Don't know how to change index name in this init node.")
+        # Change the name of the index in the comparison
```
FabioLuporini: Why are you checking between Symbol and str? str are *not* expected at this point. If there are, that should be fixed and turned into Symbols

FabioLuporini:
Apart from the various comments I made, this is not rebased on master

FabioLuporini:
This can now be rebased on master

FabioLuporini:
@doru1004  any plans for this?

miklos1:
I'm closing this now, because it has been abandoned for more than a year. If it is still relevant, please rebase and reopen.

PR #43, opened 2015-04-24
Extend expression support

FabioLuporini:
A number of features with this PR:
- Allow optimisation of generic tensor expressions (in firedrake language: we can optimise 0 and 1 forms) 
- Complete refactoring of two key operations performed by the Expression Rewriter: expansion and factorisation.
- Rewrite obsolete code with higher quality code
- Drop useless stuff
- Minor fixes
 
Must land with ffc PR 74 and pyop2 PR 437

### DIFF
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -937,9 +937,9 @@ def c_sym(const):
 
 
 def c_for(var, to, code, pragma="#pragma coffee itspace", init=None):
-    i = c_sym(var)
-    init = init or c_sym(0)
-    end = c_sym(to)
+    i = Symbol(var)
+    init = init or Symbol(0)
+    end = Symbol(to)
     if type(code) == str:
         code = FlatBlock(code)
     elif type(code) == list:
@@ -947,7 +947,7 @@ def c_for(var, to, code, pragma="#pragma coffee itspace", init=None):
     elif type(code) is not Block:
         code = Block([code], open_scope=True)
     return Block(
-        [For(Decl("int", i, init), Less(i, end), Incr(i, c_sym(1)),
+        [For(Decl("int", i, init), Less(i, end), Incr(i, Symbol(1)),
              code, pragma)], open_scope=True)
 
 
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -34,7 +34,12 @@
 """This file contains the hierarchy of classes that implement a kernel's
 Abstract Syntax Tree (AST)."""
 
-
+try:
+    from collections import OrderedDict
+# OrderedDict was added in Python 2.7. Earlier versions can use ordereddict
+# from PyPI
+except ImportError:
+    from ordereddict import OrderedDict
 from copy import deepcopy as dcopy
 
 # Utilities for simple exprs and commands
@@ -667,6 +672,11 @@ class For(Statement):
     @body.setter
     def body(self, new_body):
         self.children[0].children = new_body
+
+    @staticmethod
+    def fromloops(loops):
+        loops_dims = [l.dim for l in loops]
+        return OrderedDict(zip(loops_dims, loops))
 
     def gencode(self, not_scope=False):
         return "\n".join(self.pragma) + "\n" + for_loop(self.init.gencode(True),
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -638,7 +638,7 @@ class For(Statement):
         self.incr = incr
 
     @property
-    def itvar(self):
+    def dim(self):
         if isinstance(self.init, Decl):
             return self.init.sym.symbol
         elif isinstance(self.init, Assign):
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -565,8 +565,18 @@ class Decl(Statement, Perfect):
 
     @property
     def is_const(self):
-        """Return True if the declaration is a constant."""
+        """Return True if the declared symbol is constant"""
         return 'const' in self.qual
+
+    @property
+    def is_static(self):
+        """Return True if the declared symbol is static"""
+        return 'static' in self.qual
+
+    @property
+    def is_static_const(self):
+        """Return True if the declared symbol is static and constant"""
+        return self.is_static and self.is_const
 
     @property
     def scope(self):
@@ -580,15 +590,8 @@ class Decl(Statement, Perfect):
             raise RuntimeError("Only %s are valid scopes" % Scope._scopes)
         self._scope = val
 
-    def gencode(self, not_scope=False):
-        if isinstance(self.init, EmptyStatement):
-            return decl(spacer(self.qual), self.typ, self.sym.gencode(),
-                        spacer(self.attr)) + semicolon(not_scope)
-        else:
-            return decl_init(spacer(self.qual), self.typ, self.sym.gencode(),
-                             spacer(self.attr), self.init.gencode()) + semicolon(not_scope)
-
-    def get_nonzero_columns(self):
+    @property
+    def nonzero(self):
         """If the declared array:
 
         * is a bi-dimensional array,
@@ -601,6 +604,14 @@ class Decl(Statement, Perfect):
             return self.init.nonzero_bounds
         else:
             return ()
+
+    def gencode(self, not_scope=False):
+        if isinstance(self.init, EmptyStatement):
+            return decl(spacer(self.qual), self.typ, self.sym.gencode(),
+                        spacer(self.attr)) + semicolon(not_scope)
+        else:
+            return decl_init(spacer(self.qual), self.typ, self.sym.gencode(),
+                             spacer(self.attr), self.init.gencode()) + semicolon(not_scope)
 
 
 class Block(Statement):
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -150,8 +150,11 @@ class BinExpr(Expr):
         by the classic ``deepcopy`` method, is ignored."""
         return self.__class__(dcopy(self.children[0]), dcopy(self.children[1]))
 
-    def gencode(self, not_scope=True):
-        return (" "+self.op+" ").join([n.gencode(not_scope) for n in self.children])
+    def gencode(self, not_scope=True, parent=None):
+        subtree = (" "+self.op+" ").join([n.gencode(not_scope, self) for n in self.children])
+        if parent and not isinstance(parent, (Par, self.__class__)):
+            return wrap(subtree)
+        return subtree
 
     @property
     def left(self):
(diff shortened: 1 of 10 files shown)


coffee/base.py
```diff
@@ -384,7 +404,7 @@ class AVXSub(Sub):
 
     def gencode(self, not_scope=True):
         op1, op2 = self.children
-        return "_mm256_add_pd (%s, %s)" % (op1.gencode(), op2.gencode())
+        return "_mm256_sub_pd (%s, %s)" % (op1.gencode(), op2.gencode())
```
kynan: Suppose that was a bug?

FabioLuporini: That really was

coffee/optimizer.py
```diff
@@ -82,32 +71,26 @@ def rewrite(self, level):
...
-                                    loop-invariant code motion to move invariant \
-                                    sub-expressions exposed by factorization
-                      * level == 3: level 2 + avoid computing zero-columns
-                      * level == 4: level 3 + precomputation of read-only expressions \
-                                    out of the loop nest
+                      * level == 1: performs generalized loop-invariant code motion only
+                      * level == 2: level 1; terms expansion (to expose factorization \
+                                    opportunities); terms factorization (to expose \
+                                    code motion opportunities); and a final pass of \
+                                    generalized loop-invariant code motion.
```
kynan: Are the former levels 3 and 4 now part of 2 or are they being dropped?

FabioLuporini: The had already been moved to other passes (since they are not proper "expression rewriting" passes), so this is old doc being updated

coffee/rewriter.py
```diff
@@ -131,132 +145,131 @@ def licm(self, merge_and_simplify=False, compact_tmps=False):
...
-        # Perform expansion
-        ee = ExpressionExpander(self.stmt, self.expr_info, self.hoisted,
-                                self.expr_graph, itvar_occs, itvar_exp)
-        ee.expand()
+          is evaluated, and ``cost(L') > cost(L'')``;
+
+        :param mode: multiple expansion strategies are possible, each exposing
+                     different, "hidden" opportunities for later code motion.
+
+                     * mode == 'standard': this heuristics consists of expanding \
```
kynan: I don't think you need line continuation markers here

coffee/rewriter.py
```diff
@@ -131,132 +145,131 @@ def licm(self, merge_and_simplify=False, compact_tmps=False):
...
+                     different, "hidden" opportunities for code motion.
 
-        # Create the new expression
-        new_prods = []
-        for d in to_distr.values():
-            dist, target = zip(*d)
-            target = Par(ast_c_sum(target)) if len(target) > 1 else ast_c_sum(target)
-            new_prods.append(Par(Prod(dist[0], target)))
-        self.stmt.children[1] = Par(ast_c_sum(new_prods))
+                     * mode == 'standard': this simple heuristics consists of \
```
kynan: Don't need line continuation markers here

coffee/rewriter.py
```diff
@@ -131,132 +145,131 @@ def licm(self, merge_and_simplify=False, compact_tmps=False):
...
+class ExpressionHoister(object):
 
     # Track all expressions to which LICM has been applied
     _expr_handled = []
     # Temporary variables template
     _hoisted_sym = "%(loop_dep)s_%(expr_id)d_%(round)d_%(i)d"
 
+    # Constants used by the extract method to charaterize sub-expressions
+    INV = 0  # Invariant term, hoistable if part of an invariant expression
+    KSE = 1  # Invariant expression, potentially part of larger invariant
```
kynan: What does KSE stand for?

coffee/rewriter.py
```diff
@@ -407,124 +420,109 @@ def licm(self):
...
-                # dep[-1] which they depend on, and are wrapped by a loop wl
-                # iterating along the same iteration space as dep[-1].
-                # If there's no such an outer loop, they fall in the header,
-                # provided they are within a perfect loop nest (otherwise,
-                # dependencies may be broken)
-                outermost_loop = expr_loops[0]
-                is_outermost_perfect = is_perfect_loop(outermost_loop)
+                dep = tuple(d for d in all_deps if d in self.expr_deps)
+
+                # 0) Determine where it is safe to hoist the statements
```
kynan: Might be worth documenting each case

coffee/rewriter.py
```diff
@@ -407,124 +420,109 @@ def licm(self):
...
-                    dep_block = dict_expr_loops[dep[-2]].children[0]
-                    place, wl = dep_block, dict_expr_loops[dep[-1]]
-                    next_loop = od_find_next(dict_expr_loops, dep[-2])
+                    dep_block = expr_dims_loops[dep[-2]].children[0]
+                    place, wrap_loop = dep_block, expr_dims_loops[dep[-1]]
+                    next_loop = od_find_next(expr_dims_loops, dep[-2])
 
                 # 1) Remove identical sub-expressions
-                expr = dict([(str(e), e) for e in expr]).values()
+                exprs = dict([(str(e), e) for e in exprs]).values()
```
kynan: Why filter through a `dict` rather than a `set`?

coffee/utils.py
```diff
@@ -480,24 +537,36 @@ def count_occurrences(node, key=0, read_only=False):
...
-                * ``key == 0``: a tuple (symbol name, symbol rank)
-                * ``key == 1``: the symbol name
-                * ``key == 2``: a string representation of the symbol
+    :param mode: Accepted values are ['symbol', 'symbol_id', 'symbol_str']. This
+                 parameter drives the counting and impacts the format of the
+                 returned dictionary. In particular, the keys in such dictionary
+                 will be:
+
+                * mode == 'symbol': a tuple (symbol name, symbol rank)
+                * mode == 'symbol_id': the symbol name only (a string). This \
```
kynan: No need for line continuation markers here

coffee/vectorizer.py
```diff
@@ -31,47 +31,61 @@
...
-import plan as ap
+import plan
 
 
-class LoopVectorizer(object):
+class VectStrategy():
+
+    """Supported vectorization modes."""
+
+    # Generate scalar code suitable to compiler auto-vectorization
```
kynan: Could turn these into docstrings

kynan:
From a brief skim this looks sane.

coffee/rewriter.py
FabioLuporini: I need filtering to happen on the string representation to delete duplicates, then I pick the expressions

coffee/rewriter.py
FabioLuporini: Are you sure? I remember we (actually you, long time ago) put them in an analogous case to fix the layout (see docstring of rewrite() in optimizer.py)

coffee/rewriter.py
FabioLuporini: Doing that

FabioLuporini:
Thanks Florian, I appreciate this. I've addressed your concerns reguarding documentation.

FabioLuporini:
@wence- and @miklos1 , can we land this and the corresponding (minor) ffc and pyop2 changes?

coffee/base.py
```diff
@@ -967,3 +1013,26 @@ def __eq__(self, other):
...
+
+    def __eq__(self, other):
+        return self._scope == other._scope
+
+    def __str__(self):
+        return self._scope
+
+LOCAL = Scope("LOCAL")
+EXTERNAL = Scope("EXTERNAL")
+Scope._scopes = [LOCAL, EXTERNAL]
```
wence-: Don't do this, rather only allow people to build "LOCAL", or "EXTERNAL" scope objects.

coffee/base.py
```diff
@@ -967,3 +1013,26 @@ def __eq__(self, other):
...
+
+    """An ``EXTERNAL`` scope means the /Decl/ is an argument of a kernel (i.e.,
+    when it appears in the list of declarations of a /FunDecl/ object). Otherwise,
+    a ``LOCAL`` scope indicates the /Decl/ is within the body of a kernel."""
+
+    def __init__(self, scope):
+        self._scope = scope
+
+    def __eq__(self, other):
+        return self._scope == other._scope
```
wence-: YOu can remove the necessity for this if you always use the Scope objects you construct below.

coffee/rewriter.py
```diff
@@ -131,132 +145,131 @@ def licm(self, merge_and_simplify=False, compact_tmps=False):
...
 
     # Track all expressions to which LICM has been applied
     _expr_handled = []
     # Temporary variables template
     _hoisted_sym = "%(loop_dep)s_%(expr_id)d_%(round)d_%(i)d"
 
+    # Constants used by the extract method to charaterize expressions:
+    INV = 0  # /INVariant/: expression is loop invariant
+    KSE = 1  # /KeepSEarching/: expression is potentially in a larger invariant
+    HOI = 2  # /HOIsted/: expression is *not* invariant, should be hoisted
```
wence-: Do you mean "should not be hoisted any further"?

I know you have an inordinate love of short variable names, but I always feels like descriptive named constants ought to be so.

coffee/base.py
```diff
@@ -967,3 +1013,26 @@ def __eq__(self, other):
 DEC = Access("DEC")
 IMUL = Access("IMUL")
 IDIV = Access("IDIV")
+Access._modes = [READ, WRITE, RW, INC, DEC, IMUL, IDIV]
+
+
+# Scope of a declaration ##
+
+class Scope(object):
```
miklos1: Why is this class? It doesn't do anything other than wrapping a string.

coffee/base.py
FabioLuporini: I did not get this

coffee/rewriter.py
FabioLuporini: I'll expand these

coffee/base.py
wence-: Scope("foo") -> error

coffee/base.py
FabioLuporini: Abstraction.
Incidentally, you could raise a similar comment for Access Modes in PyOP2

coffee/base.py
FabioLuporini: True

FabioLuporini: Wait, here I'm comparing strings, not the actual objects

miklos1: He's saying that if you only use the pre-constructed objects, then object identity is more than satisfactory for comparison.

FabioLuporini: Ah right

FabioLuporini:
@wence- : addressed your comments

PR #44, opened 2015-05-07
Tree visitor infrastructure

wence-:
Add tree visitor infrastructure to COFFEE and migrate current tree inspectors to new visitors.

New visitors can be created by subclassing coffee.visitor.Visitor (which see), which comes with a number of built-in "default" methods for reusing or reconstructing nodes.

Some utility visitors (for replacing symbols matching a name or checking for uniqueness of nodes in a tree) are in coffee.visitors.utilities.

Tree inspectors are in coffee.visitors.inspectors.

Additionally, this change adds `operands` and `reconstruct` methods to all base ast nodes.  This is an endeavour to allow for tree transformations without modification in place.  Note that there may be corner cases that are not handled correctly.

Code gen should be migrated to this new infrastructure but is not yet done.

### DIFF
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -63,21 +63,6 @@ ternary = lambda e, s1, s2: wrap("%s ? %s : %s" % (e, s1, s2))
 as_symbol = lambda s: s if isinstance(s, Node) else Symbol(s)
 
 
-# Meta classes for semantic decoration of AST nodes ##
-
-
-class Perfect(object):
-    """Dummy mixin class used to decorate classes which can form part
-    of a perfect loop nest."""
-    pass
-
-
-class LinAlg(object):
-    """Dummy mixin class used to decorate classes which represent linear
-    algebra operations."""
-    pass
-
-
 # Base classes of the AST ###
 
 
@@ -137,6 +122,21 @@ class Root(Node):
     def gencode(self):
         header = '// This code is generated visiting a COFFEE AST\n\n'
         return header + Node.gencode(self)
+
+
+# Meta classes for semantic decoration of AST nodes ##
+
+
+class Perfect(Node):
+    """Dummy mixin class used to decorate classes which can form part
+    of a perfect loop nest."""
+    pass
+
+
+class LinAlg(Node):
+    """Dummy mixin class used to decorate classes which represent linear
+    algebra operations."""
+    pass
 
 
 # Expressions ###
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -90,6 +90,19 @@ class Node(object):
         # Pragmas are used to attach semantical information to nodes
         self._pragma = self._format_pragma(pragma)
 
+    def reconstruct(self, *args, **kwargs):
+        """Return a new instance of this :class:`Node`.
+
+        :arg args: positional arguments to the constructor.
+        :arg kwargs: keyword arguments to the constructor.
+        """
+        return type(self)(*args, **kwargs)
+
+    def operands(self):
+        """Return the operands of this :class:`Node` as an iterable,
+        along with a :class:`dict` of keyword arguments (the pragma decorator)."""
+        return self.children, {'pragma': self.pragma}
+
     def gencode(self):
         return "\n".join([n.gencode() for n in self.children])
 
@@ -139,9 +152,11 @@ class BinExpr(Expr):
 
     """Generic binary expression."""
 
-    def __init__(self, expr1, expr2, op):
+    def __init__(self, expr1, expr2):
         super(BinExpr, self).__init__([expr1, expr2])
-        self.op = op
+
+    def reconstruct(self, expr1, expr2, **kwargs):
+        return type(self)(expr1, expr2, **kwargs)
 
     def __deepcopy__(self, memo):
         """Binary expressions always need to be copied as plain new objects,
@@ -151,7 +166,7 @@ class BinExpr(Expr):
         return self.__class__(dcopy(self.children[0]), dcopy(self.children[1]))
 
     def gencode(self, not_scope=True, parent=None):
-        subtree = (" "+self.op+" ").join([n.gencode(not_scope, self) for n in self.children])
+        subtree = (" "+type(self).op+" ").join([n.gencode(not_scope) for n in self.children])
         if parent and not isinstance(parent, (Par, self.__class__)):
             return wrap(subtree)
         return subtree
@@ -172,6 +187,12 @@ class UnaryExpr(Expr):
     def __init__(self, expr):
         super(UnaryExpr, self).__init__([expr])
 
+    def reconstruct(self, expr, **kwargs):
+        return type(self)(expr, **kwargs)
+
+    def gencode(self, not_scope=True, parent=None):
+        return "%s%s" % (type(self).op, self.children[0].gencode(not_scope)) + semicolon(not_scope)
+
     def __deepcopy__(self, memo):
         """Unary expressions always need to be copied as plain new objects,
         ignoring whether they have been copied before; that is, the ``memo``
@@ -185,10 +206,13 @@ class UnaryExpr(Expr):
 
 
 class Neg(UnaryExpr):
+    """Unary negation of an expression"""
+    op = "-"
 
-    "Unary negation of an expression"
-    def gencode(self, not_scope=False, parent=None):
-        return "-%s" % wrap(self.child.gencode(not_scope, self)) + semicolon(not_scope)
+
+class Not(UnaryExpr):
+    """Compare an expression to ``NULL`` using the operator ``!``."""
+    op = "!"
 
 
 class ArrayInit(Expr):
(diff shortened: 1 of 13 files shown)


coffee/base.py
```diff
@@ -139,9 +147,11 @@ class BinExpr(Expr):
 
     """Generic binary expression."""
 
-    def __init__(self, expr1, expr2, op):
+    def __init__(self, expr1, expr2):
         super(BinExpr, self).__init__([expr1, expr2])
-        self.op = op
+
+    def reconstruct(self, expr1, expr2, **kwargs):
```
miklos1: So we accept and forward `**kwargs` here, but not in `__init__`. Why is that?

coffee/base.py
```diff
@@ -151,7 +161,7 @@ def __deepcopy__(self, memo):
         return self.__class__(dcopy(self.children[0]), dcopy(self.children[1]))
 
     def gencode(self, not_scope=True, parent=None):
-        subtree = (" "+self.op+" ").join([n.gencode(not_scope, self) for n in self.children])
+        subtree = (" "+type(self).op+" ").join([n.gencode(not_scope) for n in self.children])
```
miklos1: I assume you use `type(self).op` because `self.op` did not work, yet on the toy examples I have tried, `self.op` works just fine. I'm interested in what cases `self.op` fails.

coffee/base.py
wence-: So all reconstruct methods have kwargs, I was mimicking the previous code for init

coffee/base.py
wence-: Oh, I was just looking at the class property because it's a class property. I didn't realise I can access class properties directly on instances

coffee/base.py
```diff
@@ -722,6 +709,12 @@ def __init__(self, switch_expr, cases):
         self.switch_expr = switch_expr
         self.cases = cases
 
+    def operands(self):
+        return [self.switch_expr, self.cases], {}
+
+    def reconstruct(self, expr, cases, **kwargs):
```
miklos1: Why are keyword arguments dropped silently?

wence-: Oversight

coffee/visitor.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Pre- and post-order traversal is implemented in a single
+    :meth:`visit` method, with the argument list for a method handler
+    indicating if it expects transformed children or not.
+
+    The two method signatures are:
+
+    .. code-block::
+
+       def visit_Foo(self, o, env, **kwargs):
```
miklos1: Why do we need the `**kwargs` in this case?

coffee/visitors/inspectors.py
```diff
@@ -0,0 +1,486 @@
...
+    are given potential unroll factors that result in no loop
+    remainders.
+    """
+    def visit_object(self, o, env):
+        return {}
+
+    def visit_Node(self, o, env, *args, **kwargs):
+        ret = {}
+        for a in args:
+            for k, v in a.iteritems():
```
miklos1: `ret.update(a)`?

coffee/visitor.py
wence-: Doc bug, you don't. In fact you don't get them.

coffee/visitors/inspectors.py
wence-: Probably

miklos1:
I'm slightly more keen on using the shorter `dict(in_loop=False, multiple_statements=False)` for top level environments instead of `Environment({}, in_loop=False, multiple_statements=False)`, although they behave the same way.

miklos1:
Looking at it now, I'm a bit wary about the practice of adopting the default environment *in the `visit` method* when no environment was specified. One might accidentally forget to forward the environment to the children in a `visit_Foo`, but that is no error: visitation silently continues with the default top-level environment which is probably wrong and likely leads to harder to find errors. I suggest not having any default values for `env`.

I see that the default environment "feature" is mostly there for external callers of the visitor, so they don't have to bother with a default top-level environment. but I think it would be better practice for external callers to use - possibly readily offered - wrapper functions, which feed the default top-level environment at the first call to the visitor. Many of these wrapper functions already exist (e.g. `inner_loops` for `FindInnerLoops`), in those cases one just has to move the default environment into the wrapper from the generic `visit` method of class.

FabioLuporini:
I only have three minor points, although not actual objections:
- It's a pity two different ways of traversing the tree are co-existing, but there's not much we can do for this at the moment
- I will need help with rebasing the branch enhance-rewriting
- I still don't quite understand the point of in-place vs out-place AST manipulation. If nodes are ensured to be unique (if they are not at the moment, we can do so anyway), what's the problem?

wence-:
Well, you can always visit the tree "by hand", which is the existing method, right, so I don't know what the point is really.

Fine, I do not think it should be too difficult.

As to in-place vs. out-of-place manipulation:

Principle of least surprise.  In-place manipulations are surprising since they affect code "far away".  This leads, in my view, to code that is hard to understand, and hard to maintain.

I don't buy the uniqueness argument.  It is certainly not true, /right now/, that asts fed to coffee contain unique nodes.  Moreover, I would argue that enforcing uniqueness actively makes some kind of transformations difficult.  I'm aware that this is not quite how the code is written at the moment, but imagine a `Symbol` node is changed to only carry the name, with shape and indexing an extra node.  Now consider some code:

```c
{
    int a;
    ....
    {
         for ( int a; ... ) {
             ...
         }
         f(a);
    }
}```

And I, for whatever reason, want to rename the iteration index in the for loop.  If I have unique symbols whenever I refer to `a` then my replacement has to map names to new symbols.  This makes renaming the iteration variable much more difficult.  I can't just drop down until I see that instance and replace it, reconstructing the parents.  I have to match on names, and check that I'm in the for loop and so forth.

Obviously, this is not how symbols work right now, but I sort of think that having:

```
# a
a = Symbol("a")
i = Symbol("i")
j = Symbol("j")
# a[3][3]
Variable(a, Shape(3, 3))
# a[i][j]
Variable(a, Index(i, j))
```
Or something similar, is a nicer model, since the "name" is distinct from the "annotations".

With in place modification as well, I always have to extremely careful if I'm making modifications that I don't accidentally change something "elsewhere": say I hoist an expression and rewrite it, I always have to remember to copy it first (in case it's used elsewhere), which is, again, fragile.

I agree that this is not the path of least resistance, but in the long term, I suspect it will be beneficial.

**Comment by FabioLuporini** [comment]
OK, I see.

**Comment by wence-** [comment]
OK, I think I have addressed comments.  And also added some tests, although they have very incomplete coverage.

**Comment by wence-** [comment]
Now maybe actually working.

I have a comment about the `Environment` class.  `__getitem__` looks things up in the parent but other dict methods currently don't, so `env.get("foo", "bar")` will return `"bar"` if the current Environment doesn't have a "foo" key (even if parents do).  This is probably surprising!

**Comment by FabioLuporini** [comment]
Don't know if you care, but you can add yourself to the list of AUTHORS.

Thanks.

**Comment by miklos1** [comment]
@wence-: I probably would not have derived `Environment` from `dict`, but rather have the dictionary for the frame as an attribute. That way no surprising behaviour is inherited, since only those functions work which are explicitly enabled. In that case we likely want a `__str__` too.

Another option is simply adding the `get` method to the right thing. However, `get` and `__getitem__` are probably not the only two things that work on dictionaries (e.g. iteration), but would work the wrong way on environments.

**Comment by miklos1** [comment]
More on environments: I also think that probably `__getitem__` is the only right way to query an environment. Giving a default value in a `get`-like manner is superfluous to having the top-level environment set up properly.

**Comment by wence-** [comment]
That makes sense.  I note it's not ready to go yet since my most recent round of cleanups broke things again.

**Comment by wence-** [comment]
OK, I've migrated environments to just inherit from object and implemented `__getitem__`, `__setitem__`, `__str__` and `__repr__`.

**Comment by miklos1** [comment]
I like the new `Environment`, `__str__` and `__repr__` are nice, too.
However, `__setitem__` seems to go somewhat against the purpose of the environment, which serve for a side-effect-free propagation of information downwards in the tree. Ideally, one just extends it with another frame (which may shadow values), and queries values.

**Comment by wence-** [comment]
That's a reasonable point, I've removed it.  Also, more tests.

**Comment by wence-** [comment]
OK, I think this is now good to go, any final comments before merge?

File: coffee/visitors/inspectors.py [hunk 6]
```diff
@@ -0,0 +1,540 @@
...
+    Each list entry describes a loop nest with the outer-most loop
+    first.  Each entry therein is a tuple (loop_node, node_parent).
+
+    By default the top-level call to visit will record a node_parent
+    of None for the visited Node.  To provide one, pass an
+    environment in to the visitor::
+
+    .. code-block::
+
+       v.visit(node, {node_parent: parent})
```
miklos1: `v.visit(node, {'node_parent': parent})`? (Quotes around `node_parent`.)
You probably want replace all, this is not the only such documentation comment.

miklos1:
I just spotted a possible docstring bug.
Otherwise, I think this is now good to go.

PR #46, opened 2015-05-11
Travisify

wence-:
Enable build testing via travis.


FabioLuporini:
Can go. Why have you added "E731,E402" ?

wence-:
Because I don't want to remove all the one-line lambdas and non top-level module imports

PR #51, opened 2015-05-19
Enhance rewriting, bug fixing

FabioLuporini:
- bug fixing (mostly concerning expression rewriting)
- infrastructure for novel rewriting strategies
- code refactoring, including rewriting ancient pieces of code

this depends on PR 77 in FFC

fixes issues 45, 49, 50

### DIFF
--- a/coffee/ast_analyzer.py
+++ b/coffee/ast_analyzer.py
@@ -133,8 +133,17 @@ class ExpressionGraph(object):
 
     """Track read-after-write dependencies between symbols."""
 
-    def __init__(self):
+    def __init__(self, node):
+        """Initialize the ExpressionGraph.
+
+        :param node: root of the AST visited to initialize the ExpressionGraph.
+        """
         self.deps = nx.DiGraph()
+        search = visit(node, search=(Assign, Incr, Decr, IMul, IDiv))['search']
+        for type, nodes in search.items():
+            nodes = [n for n in nodes if not isinstance(n.children[1], int)]
+            for n in nodes:
+                self.add_dependency(*n.children)
 
     def add_dependency(self, sym, expr):
         """Add dependency between ``sym`` and symbols appearing in ``expr``."""
--- a/coffee/ast_analyzer.py
+++ b/coffee/ast_analyzer.py
@@ -103,24 +103,6 @@ class StmtTracker(OrderedDict):
             if sym_info.loop == loop_a:
                 self.update_stmt(sym, **{'loop': loop_b})
 
-    def delete_hoisted(self, sym):
-        """Remove the hoisted statement whose temporary is symbol ``sym``.
-        The removal is total: it is deleted from the loop nest in which it is
-        computed, as well as from the dictionary."""
-
-        decl = self[sym].decl
-        place = self[sym].place
-        loop = self[sym].loop
-
-        place.children.remove(decl)
-        to_remove = None
-        for stmt in loop.children[0].children:
-            if stmt.children[0].symbol == sym:
-                to_remove = stmt
-        loop.children[0].children.remove(to_remove)
-
-        self.pop(sym)
-
     @property
     def expr(self, sym):
         return self[sym].expr if self.get(sym) else None
--- a/coffee/ast_analyzer.py
+++ b/coffee/ast_analyzer.py
@@ -41,6 +41,7 @@ except ImportError:
 import networkx as nx
 
 from base import *
+from utils import visit
 
 
 class StmtInfo():
@@ -135,36 +136,38 @@ class ExpressionGraph(object):
     def __init__(self):
         self.deps = nx.DiGraph()
 
-    def add_dependency(self, sym, expr, self_loop):
-        """Extract symbols from ``expr`` and create a read-after-write dependency
-        with ``sym``. If ``sym`` already has a dependency, then ``sym`` has a
-        self dependency on itself."""
+    def add_dependency(self, sym, expr):
+        """Add dependency between ``sym`` and symbols appearing in ``expr``."""
+        expr_symbols = visit(expr, search=Symbol)['search'][Symbol]
+        for es in expr_symbols:
+            self.deps.add_edge(sym.symbol, es.symbol)
 
-        def extract_syms(sym, node, deps):
-            if isinstance(node, Symbol):
-                deps.add_edge(sym, node.symbol)
-            else:
-                for n in node.children:
-                    extract_syms(sym, n, deps)
+    def is_read(self, expr, target_sym=None):
+        """Return True if any symbols in ``expr`` is read by ``target_sym``,
+        False otherwise. If ``target_sym`` is None, Return True if any symbols
+        in ``expr`` are read by at least one symbol, False otherwise."""
+        input_syms = visit(expr, search=Symbol)['search'][Symbol]
+        for s in input_syms:
+            if s.symbol not in self.deps:
+                continue
+            elif not target_sym:
+                if zip(*self.deps.in_edges(s.symbol)):
+                    return True
+            elif nx.has_path(self.deps, target_sym.symbol, s.symbol):
+                return True
+        return False
 
-        sym = sym.symbol
-        # Add self-dependency
-        if self_loop:
-            self.deps.add_edge(sym, sym)
-        extract_syms(sym, expr, self.deps)
-
-    def has_dep(self, sym, target_sym=None):
-        """If ``target_sym`` is not provided, return True if ``sym`` has a
-        read-after-write dependency with some other symbols. This is the case if
-        ``sym`` has either a self dependency or at least one input edge, meaning
-        that other symbols depend on it.
-        Otherwise, if ``target_sym`` is not None, return True if ``sym`` has a
-        read-after-write dependency on it, i.e. if there is an edge from
-        ``target_sym`` to ``sym``."""
-
-        sym = sym.symbol
-        if not target_sym:
-            return sym in self.deps and zip(*self.deps.in_edges(sym))
-        else:
-            target_sym = target_sym.symbol
-            return sym in self.deps and self.deps.has_edge(sym, target_sym)
+    def is_written(self, expr, target_sym=None):
+        """Return True if any symbols in ``expr`` is written by ``target_sym``,
+        False otherwise. If ``target_sym`` is None, Return True if any symbols
+        in ``expr`` are written by at least one symbol, False otherwise."""
+        input_syms = visit(expr, search=Symbol)['search'][Symbol]
+        for s in input_syms:
+            if s.symbol not in self.deps:
+                continue
+            elif not target_sym:
+                if zip(*self.deps.out_edges(s.symbol)):
+                    return True
+            elif nx.has_path(self.deps, s.symbol, target_sym.symbol):
+                return True
+        return False
--- a/coffee/ast_analyzer.py
+++ b/coffee/ast_analyzer.py
@@ -41,7 +41,7 @@ except ImportError:
 import networkx as nx
 
 from base import *
-from utils import visit
+from coffee.visitors import FindInstances
 
 
 class StmtInfo():
@@ -139,15 +139,15 @@ class ExpressionGraph(object):
         :param node: root of the AST visited to initialize the ExpressionGraph.
         """
         self.deps = nx.DiGraph()
-        search = visit(node, search=Writer)['search']
-        for type, nodes in search.items():
+        writes = FindInstances(Writer).visit(node)
+        for type, nodes in writes.items():
             nodes = [n for n in nodes if not isinstance(n.children[1], int)]
             for n in nodes:
                 self.add_dependency(*n.children)
 
     def add_dependency(self, sym, expr):
         """Add dependency between ``sym`` and symbols appearing in ``expr``."""
-        expr_symbols = visit(expr, search=Symbol)['search'][Symbol]
+        expr_symbols = FindInstances(Symbol).visit(expr)[Symbol]
         for es in expr_symbols:
             self.deps.add_edge(sym.symbol, es.symbol)
 
@@ -155,7 +155,7 @@ class ExpressionGraph(object):
         """Return True if any symbols in ``expr`` is read by ``target_sym``,
         False otherwise. If ``target_sym`` is None, Return True if any symbols
         in ``expr`` are read by at least one symbol, False otherwise."""
-        input_syms = visit(expr, search=Symbol)['search'][Symbol]
+        input_syms = FindInstances(Symbol).visit(expr)[Symbol]
         for s in input_syms:
             if s.symbol not in self.deps:
                 continue
@@ -170,7 +170,7 @@ class ExpressionGraph(object):
         """Return True if any symbols in ``expr`` is written by ``target_sym``,
         False otherwise. If ``target_sym`` is None, Return True if any symbols
         in ``expr`` are written by at least one symbol, False otherwise."""
-        input_syms = visit(expr, search=Symbol)['search'][Symbol]
+        input_syms = FindInstances(Symbol).visit(expr)[Symbol]
         for s in input_syms:
             if s.symbol not in self.deps:
                 continue
--- a/coffee/ast_analyzer.py
+++ b/coffee/ast_analyzer.py
@@ -47,7 +47,7 @@ from coffee.visitors import FindInstances
 class StmtInfo():
     """Simple container class defining ``StmtTracker`` values."""
 
-    INFO = ['expr', 'decl', 'loop', 'place']
+    INFO = ['stmt', 'decl', 'loop', 'place']
 
     def __init__(self, **kwargs):
         for k, v in kwargs.iteritems():
@@ -61,17 +61,16 @@ class StmtTracker(OrderedDict):
 
     Each key in the dictionary is a string representing a symbol. As such,
     StmtTracker can be used only in SSA scopes. Each entry in the dictionary
-    is tuple containing information about the symbol: ::
+    is a 4-tuple containing information about the symbol: ::
 
-        (expression, declaration, closest_for, place)
+        (statement, declaration, closest_for, place)
 
     whose semantics is, respectively, as follows:
 
-        * The AST root node of the right-hand side of the statement whose
-          left-hand side is ``sym``
+        * The AST node whose ``str(lvalue)`` is used as dictionary key
         * The AST node of the symbol declaration
         * The AST node of the closest loop enclosing the statement
-        * The parent block of the loop
+        * The parent of the closest loop
     """
 
     def __setitem__(self, key, value):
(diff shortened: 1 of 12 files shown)


coffee/base.py
```diff
@@ -221,20 +217,61 @@ class ArrayInit(Expr):
...
+            raise TypeError
+        self._values = val
+
+    def _formatter(self, v):
+        """Format a float into a string, showing up to ``precision`` decimal digits.
+        This function is partly extracted from the open_source "FFC: the FEniCS Form
+        Compiler", freely accessible at https://bitbucket.org/fenics-project/ffc."""
+        f = "%%.%dg" % self.precision
+        f_int  = "%%.%df" % 1
+        eps = eval("1e-%s" % self.precision)
```
wence-: AKA `eps = 10.0**(-self.precision)`

coffee/base.py
```diff
@@ -221,20 +217,61 @@ class ArrayInit(Expr):
...
+        return f_int % v if abs(v - round(v, 1)) < eps else f % v
```
wence-: I'm confused by this function, why is it not just as simple as `return ("%%.%dg" % self.precision) % v`?

coffee/rewriter.py
```diff
@@ -253,6 +210,179 @@ def factorize(self, mode='standard'):
...
+            a*b[i]*c[i][j]*d
+
+        are reordered as ::
+
+            a*d*b[i]*c[i][j]
+
+        This as achieved by reorganizing the AST of the expression.
+        """
+
+        def _reassociate(node, parent):
```
miklos1: Shouldn't this be a visitor?

coffee/base.py
```diff
@@ -243,25 +280,27 @@ class ColSparseArrayInit(ArrayInit):
...
-        return type(self)(values, nonzero_bounds, numpy_values, **kwargs)
+    def reconstruct(self, values, precision, nonzero_bounds, **kwargs):
+        return type(self)(values, precision, nonzero_bounds, **kwargs)
 
     def operands(self):
-        return [self.values, self.nonzero_bounds, self.numpy_values], {}
-
-    def gencode(self, not_scope=True, parent=None):
-        return self.values
+        return [self.values, self.nonzero_bounds], {}
```
wence-: This can't be the correct argument list to return from operands, because where has the precision gone?

coffee/base.py
```diff
@@ -221,20 +217,61 @@ class ArrayInit(Expr):
...
         self.values = values
+        self.precision = precision or ArrayInit._default_precision
 
-    def reconstruct(self, values, **kwargs):
-        return type(self)(values, **kwargs)
+    def reconstruct(self, values, precision, **kwargs):
+        return type(self)(values, precision, **kwargs)
 
     def operands(self):
         return [self.values], {}
```
wence-: You need to return the precision as well, otherwise it's lost when reconstructing via operands.

coffee/rewriter.py
```diff
@@ -286,102 +412,92 @@ def __init__(self, stmt, expr_info, header, decls, hoisted, expr_graph):
...
+            if isinstance(node, Symbol):
+                should_extract = False
+            elif hoist_out_domain:
+                if not set(dep).issubset(set(self.expr_info.out_domain_dims)):
+                    should_extract = False
+            if should_extract:
+                dep_subexprs[dep].append(node)
+            return should_extract
+
+        def __extract(node, dep_subexprs):
```
miklos1: Shouldn't this be a visitor?

coffee/ast_analyzer.py
```diff
@@ -150,39 +132,50 @@ class ExpressionGraph(object):
...
-            return sym in self.deps and zip(*self.deps.in_edges(sym))
-        else:
-            target_sym = target_sym.symbol
-            return sym in self.deps and self.deps.has_edge(sym, target_sym)
+        :param node: root of the AST visited to initialize the ExpressionGraph.
+        """
+        self.deps = nx.DiGraph()
+        writes = FindInstances(Writer).visit(node)
+        for type, nodes in writes.items():
+            nodes = [n for n in nodes if not isinstance(n.children[1], int)]
```
wence-: Why do nodes whose second child is an int not count towards any dependencies.

coffee/rewriter.py
```diff
@@ -607,60 +708,63 @@ def _hoist(self, expansion, exp, grp):
...
         # Update tracked information
         self.expanded_decls[decl.sym.symbol] = decl
-        self.hoisted[hoisted_exp.symbol] = (expr, decl, hoisted_loop, hoisted_place)
-        self.expr_graph.add_dependency(hoisted_exp, expr, 0)
+        self.hoisted[hoisted_exp.symbol] = (stmt, decl, hoisted_loop, hoisted_place)
+        self.expr_graph.add_dependency(hoisted_exp, expr)
+        self.cache[cache_key] = hoisted_exp
         return {exp: hoisted_exp}
 
     def _expand(self, node, parent):
```
miklos1: Visitor here?

coffee/visitors/inspectors.py
```diff
@@ -337,14 +337,28 @@ class SymbolDependencies(Visitor):
...
-    empty) loop list the symbol depends on, and maps symbol names to
-    WRITE access.
+    The returned dict contains maps from nodes to a (possibly
+    empty) loop list the symbol depends on.
     """
 
     def visit_Symbol(self, o, env):
-        # Empty loop nest for starters
-        return {o: []}
+        try:
```
miklos1: Can you incorporate write=False into the default top-level environment?

coffee/visitors/inspectors.py
```diff
@@ -337,14 +337,28 @@ class SymbolDependencies(Visitor):
...
+            write = env["write"]
+        except KeyError:
+            write = False
+        try:
```
miklos1: Can you incorporate loop_nest=[] into the default top-level environment?

coffee/visitors/inspectors.py
```diff
@@ -368,38 +382,49 @@ def visit_Decl(self, o, env):
...
+        ret.update(lvalue)
         for r in o.children[1:]:
             d = self.visit(r, env=env)
             ret.update(d)
         return ret
 
     def visit_For(self, o, env):
+        try:
+            loop_nest = env["loop_nest"]
+        except KeyError:
```
miklos1: Can you incorporate loop_nest=[] into the default top-level environment?

coffee/visitors/inspectors.py
```diff
@@ -504,15 +531,15 @@ class FindInstances(Visitor):
...
-          tree.
-    :arg stop_when_found: optional, don't traverse the children of
-          matching types.
+    :arg types: list of types or single type to search for in the tree.
+    :arg stop_when_found: optional, don't traverse the children of matching types.
+    :arg with_parent: optional, track also the parent of the matching type.
     """
 
-    def __init__(self, types, stop_when_found=False):
+    def __init__(self, types, stop_when_found=False, with_parent=False):
```
miklos1: Wouldn't it be simpler to just collect the parent information, and then drop it as a post-processing step on the resulting dictionary?

coffee/rewriter.py
```diff
@@ -253,6 +210,179 @@ def factorize(self, mode='standard'):
...
+                for n, p in other_nodes:
+                    _reassociate(n, p)
+                # Create the reassociated product and modify the original AST
+                children = zip(*other_nodes)[0] if other_nodes else ()
+                children += tuple(reorder(zip(*symbols)[1] if symbols else ()))
+                reassociated_node = ast_make_expr(Prod, children)
+                parent.children[parent.children.index(node)] = reassociated_node
+
+            else:
+                warning('Unexpect node of type %s while reassociating', typ(node))
```
wence-: What happens in this case?  Does it just not matter?

coffee/visitors/utilities.py
```diff
@@ -78,3 +81,106 @@ def visit_object(self, o, env):
...
+    def visit_Node(self, o, env):
+        ret = {}
+        for n in o.children:
+            ret.update(self.visit(n, env=env))
+        return ret
+
+    def visit_For(self, o, env):
+        try:
+            loop_nest = env["loop_nest"]
+        except KeyError:
```
miklos1: Fix default top-level environment?

coffee/visitors/utilities.py
```diff
@@ -78,3 +81,106 @@ def visit_object(self, o, env):
...
+            loop_nest = []
+        new_env = Environment(env, loop_nest=loop_nest + [o])
+        return self.visit(o.body, env=new_env)
+
+    def visit_Writer(self, o, env):
+        try:
+            loop_nest = env["loop_nest"]
+        except KeyError:
```
miklos1: Fix default top-level environment?

coffee/visitors/utilities.py
```diff
@@ -78,3 +81,106 @@ def visit_object(self, o, env):
...
+            return
+        return self.mapper[o.__class__](*args)
+
+    def visit_Par(self, o, env):
+        return self.visit(o.child, env)
+
+    def visit_Symbol(self, o, env):
+        try:
+            point = env["point"]
+        except KeyError:
```
miklos1: Fix default top-level environment?

coffee/rewriter.py
```diff
@@ -253,6 +210,179 @@ def factorize(self, mode='standard'):
...
+        """
+        # Aliases
+        stmt, expr_info = self.stmt, self.expr_info
+
+        # Division replacement
+        divisions = FindInstances(Div).visit(stmt.children[1])[Div]
+        to_replace = {}
+        for i in divisions:
+            if isinstance(i.right, Symbol) and isinstance(i.right.symbol, float):
+                to_replace[i] = Prod(i.left, Div(1, i.right.symbol))
```
wence-: Is there a reason you don't go straight to `Prod(i.left, 1/i.right.symbol)`?

coffee/rewriter.py
```diff
@@ -253,6 +210,179 @@ def factorize(self, mode='standard'):
...
+        ast_replace(stmt.children[1], to_replace, copy=True, mode='symbol')
+
+        ###
+
+        # Loop reduction
+        if not stmt.__class__ in [Incr, Decr, IMul, IDiv]:
```
wence-: idiomatically:

`if not isinstance(stmt, (Incr, Decr, IMul, IDiv))`

wence-: Or do you have subclasses of these which you don't want to accidentally catch?

coffee/base.py
FabioLuporini: I just took the code from FFC, I didn't think about that

coffee/rewriter.py
FabioLuporini: Yes, but I wrote it before Lawrence wrote the infrastructure, so maybe one day we can turn it into a visitor. Not gonna happen now.

coffee/base.py
FabioLuporini: I forgot the precision, will add it

coffee/base.py
FabioLuporini: OK

coffee/rewriter.py
FabioLuporini: Yes, but I wrote it before Lawrence wrote the infrastructure, so maybe one day we can turn it into a visitor. Not gonna happen now.

coffee/ast_analyzer.py
FabioLuporini: At the moment, I can't even think of a case when a child could be an int...will let you know

coffee/rewriter.py
FabioLuporini: Yes, but I wrote it before Lawrence wrote the infrastructure, so maybe one day we can turn it into a visitor. Not gonna happen now.

coffee/visitors/inspectors.py
FabioLuporini: see below

coffee/visitors/inspectors.py
FabioLuporini: see below

coffee/visitors/inspectors.py
FabioLuporini: Ah yes! You told me about this. Yes, will do that.

coffee/visitors/inspectors.py
FabioLuporini: maybe

coffee/rewriter.py
FabioLuporini: maybe there's some fancy code that uses nodes which I'm not expecting here, I don't know

coffee/rewriter.py
FabioLuporini: I don't know if double precision arithmetic in python is identical to that of ANSI C compilers?

coffee/rewriter.py
FabioLuporini: I was probably drunk when I wrote that, I have no idea of why I used __class__, will change it

coffee/optimizer.py
```diff
@@ -73,26 +71,115 @@ def rewrite(self, level):
...
             ew = ExpressionRewriter(stmt, expr_info, self.decls, self.header,
                                     self.hoisted, self.expr_graph)
-            if level > 0:
+            # 1) Rewrite the expressions
+            if expr_info.dimension in [0, 1]:
+                if mode in [1, 2, 3]:
+                    ew.licm(hoist_out_domain=True)
+                continue
+
+            if mode == 1:
```
kynan: We only get here if dimension >= 2?

coffee/rewriter.py
```diff
@@ -253,6 +210,179 @@ def factorize(self, mode='standard'):
...
+
+        # Full unroll any unrollable, injectable loop
+        for nest in injectable_nests:
+            to_unroll = [(l, p) for l, p in nest if l not in self.expr_info.loops]
+            nest_writers = FindInstances(Writer).visit(to_unroll[0][0])
+            for op, stmts in nest_writers.items():
+                if op in [Assign, IMul, IDiv]:
+                    # Unroll is unsafe, skip
+                    continue
+                # At this point, must be /op = Incr/
```
kynan: So you could assert that?

coffee/optimizer.py
FabioLuporini: yes

coffee/rewriter.py
FabioLuporini: yes, will do

coffee/base.py
FabioLuporini: Maybe this is to avoid returning "integer strings", such as "1" (instead you want "1.0")? I'm just guessing

wence-: What is it you want to happen when formatting an array to a certain precision?  Not, "what did the previous code that I don't understand do?"

FabioLuporini: I want:
- values to always be in float format (1.0 etc, not 1)
- values to show up to ``self.precision`` digits after the '.' , where up to means that if you have 1.0000, you only show 1.0

wence-: `np.array2string(array, precision=self.precision, supress_small=True, separator=", ").replace("[", "{").replace("]", "}")`

coffee/visitors/utilities.py
FabioLuporini: Done

FabioLuporini:
I think I addressed the main observations from Lawrence and Miklos. I think we should land this ASAP

wence-:
FWIW, this makes some code I have /run/ again with coffee on.  However, it significantly changes the convergence behaviour.

This is a simple hyper-elastic model of a rubber block undergoing compression under its own weight plus a traction force on the top.

With coffee off, I converge the residual to 1e-14 in 4 nonlinear iterations.  With O2 on, it takes 6 iterations and I'm converged to 1e-8.  With O4 it takes 26 iterations!  I got bored waiting for the kernel transformation to complete with O3.

```python
from firedrake import *
parameters["coffee"]["O2"] = False
# parameters["coffee"]["O2"] = True
mesh = RectangleMesh(2, 2, 10, 10)

V = VectorFunctionSpace(mesh, "Lagrange", 1)


bc_bottom = DirichletBC(V, Constant((0, 0)), 3)

v  = TestFunction(V)
u = Function(V)
T  = Constant((-0.25, 0))  # Traction force on the boundary
B  = Constant((0, -0.25))

# Kinematics
d = u.geometric_dimension()
I = Identity(d)             # Identity tensor
F = I + grad(u)             # Deformation gradient
C = F.T*F                   # Right Cauchy-Green tensor

# Invariants of deformation tensors
Ic = tr(C)
J  = det(F)

# Elasticity parameters
mu = Constant(6.3)
lmbda = Constant(10.0)

# Stored strain energy density (compressible neo-Hookean model)
psi = (mu/2)*(Ic - 3) - mu*ln(J) + (lmbda/2)*(ln(J))**2

# Total potential energy
Pi = psi*dx - dot(T, u)*ds(4) - dot(B, u)*dx

# Compute first variation of Pi (directional derivative about u in the direction of v)
F = derivative(Pi, u, v)

solve(F == 0, u, bcs=bc_bottom, solver_parameters={"snes_monitor": True,
                                                   "pc_type": "lu",
                                                   "ksp_monitor": True})
```

On the plus side, on larger problems, with coffee on each individual step is noticeably faster :).

wence-:
If I remove the `ew.factorize()` step from O2, this difference goes away.  If I /just/ do factorization, the problem doesn't converge at all.

FabioLuporini:
O3 and O4 still have to be tested for performance, for O2, that means numerical values are different? that's suspicious

miklos1:
I think you also need to make flake8 happy - we don't like to see a failed build.

FabioLuporini:
can you put the code somewhere?

wence-:
It's in my comment, no?

FabioLuporini:
lolz, sure

wence-:
I note the tests need updating for the new default environment setup. Also, flake8

coffee/visitors/utilities.py
miklos1: As far as I can see, this is fixed at all other occasions except for this one.

FabioLuporini: No, this is somewhat "special", because it's not a top-level environment; rather, it's pushed (and successively popped) only by Writer nodes

miklos1: I don't quite understand this. Can you write a more detailed explanation in the source code as a comment?

FabioLuporini: fine

FabioLuporini:
flake8 fixed, factorization fixed (it was breaking because in the last rebase I cut away accidentally some changes...), last comments from Miklos cleared. Any other objections?

FabioLuporini:
Ah! Also updated the tests now.

@wence- : your code converges in 4 iterations now with O2 on my machine

miklos1:
I think it is now good to go, assuming it does not break any PyOP2/Firedrake tests either.

FabioLuporini:
I'll leave the merging to you -- just remember the dependency on the ffc branch (they must land together)

PR #52, opened 2015-05-27
More faster

wence-:
Rework the visitor infrastructure to not be horrifically slow.  In particular, handlers must now always take care to visit children and the `*args` and `**kwargs` may be used to pass state up and down the tree.  This means we no longer need an `Environment` object.  The state passing is used in a number of places to pass a container used to combine results into (rather than consing and merging).  Visitor classes can define a `default_retval` classmethod to ease the user thereof in constructing an appropriate empty container.

This makes bendy slightly more bearable in terms of performance, although runtime still blows up linearly in the size of the ast we're transforming (bendy forms are ballpark 20x bigger and take ballpark 20x longer?).  Also makes non-bendy coffee transformations run faster.

Comments welcome,  everything seems to work.


kynan:
Looks very sensible to me.

FabioLuporini:
"This makes bendy slightly more bearable in terms of performance, although runtime still blows up linearly in the size of the ast we're transforming" you mean actual runtime or ast-processing time?

coffee/rewriter.py
```diff
@@ -143,7 +143,7 @@ def expand(self, mode='standard'):
                      * mode == 'full': expansion is performed aggressively without \
                                        any specific restrictions.
         """
-        symbols = FindInstances(Symbol).visit(self.stmt.children[1])[Symbol]
+        symbols = FindInstances(Symbol).visit(self.stmt.children[1], ret=FindInstances.default_retval())[Symbol]
```
FabioLuporini: I always try to have lines shorter than 90 cols (*rarely* I get to ~93-94), so we should stick to it

FabioLuporini:
Take a line like this
``FindInstances(Writer).visit(node, ret=FindInstances.default_retval())``
can we not impose da default_reval is a "real" default thing, so we don't have to specify it unless we want something different than default?

FabioLuporini:
The rest looks reasonable to me

FabioLuporini:
prod

wence-:
As to the default return value thing.  I am wary of magically creating one at the top level (there are potentially some slightly subtle interactions), I would prefer to pass it explicitly but let the object/class provide a uniform way of getting one.  Recall we had this discussion the first time round I think.

FabioLuporini:
Fine, shrink the lines and then you can hit merge. Before landing the other PR (branch update-zero-tracking or something) we'll have to update the new visitors (or visitor calls, I don't quite remember) in there to the new ones

wence-:
Done, go ahead and merge.

FabioLuporini:
Merged.

PR #54, opened 2015-06-02
Fix dep tracking

wence-:


FabioLuporini:
We should land this

kynan:
Looks good to me.

PR #56, opened 2015-06-10
Add conda recipe and binstar-build file

gmarkall:
The conda recipe enables building of conda packages, which can be done on a local machine or on Binstar. The testing regime is inspired by the existing .travis.yml - conda will run the tests on the package before declaring success.

The second part of this PR is the binstar-build file, which executes the conda build on a binstar worker, and then uploads the generated package to the firedrakeproject's channel - this can be installed by a conda user who runs `conda install -c firedrakeproject coffee`. (It is also possible to add firedrakeproject to their default conda channels so that the `-c firedrakeproject` argument is unnecessary - that is down to the user's choice).

I'm not sure I've done the right thing for the version number here, as I'm not sure what the convention for it should be for coffee. One common thing to do for packages that are builds of a development branch and not a release is to use the tag from git describe as the version number, then use the number of commits since that tag for the build number - this guarantees that the package name is monotonically increasing for each subsequent revision of the package, and therefore conda will always be able to find the latest build. I couldn't do this here because there are no tags, but if a tag were added I could update this PR to use that method. If another version/build number scheme is preferable, please let me know and I'll update the PR to use it.

### DIFF
--- a/.binstar.yml
+++ b/.binstar.yml
@@ -1,0 +1,13 @@
+package: coffee
+user: firedrakeproject
+
+platform:
+  - linux-64
+
+engine:
+  - python=2.7
+
+script:
+  - conda build -q condarecipe
+
+build_targets: conda
--- a/condarecipe/build.sh
+++ b/condarecipe/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install
--- a/condarecipe/meta.yaml
+++ b/condarecipe/meta.yaml
@@ -1,9 +1,12 @@
 package:
   name: coffee
-  version: 0.1.0
+  version: {{ environ.get('GIT_DESCRIBE_TAG','') }}
 
 source:
   path: ..
+
+build:
+  number: {{ environ.get('GIT_DESCRIBE_NUMBER', 0) }}
 
 requirements:
   build:
--- a/condarecipe/meta.yaml
+++ b/condarecipe/meta.yaml
@@ -1,0 +1,29 @@
+package:
+  name: coffee
+  version: 0.1.0
+
+source:
+  path: ..
+
+requirements:
+  build:
+    - python
+    - networkx
+
+  run:
+    - python
+    - numpy
+    - networkx
+
+test:
+  requires:
+    - pytest
+    - flake8
+  commands:
+    - py.test {{ os.path.join(environ.get('SRC_DIR'), 'tests') }} -v
+    - flake8 {{ environ.get('SRC_DIR') }}
+
+about:
+  home: http://www.firedrakeproject.org
+  license: BSD 3-clause
+  summary: COFFEE - COmpiler For Fast Expression Evaluation


kynan:
I think this is good to go. Agree with your thoughts on build numbers. We can always change it once there is a tagged "release" of COFFEE.

gmarkall:
The trouble with it as it is is that each subsequent build will overwrite the previous one - the version of the package 0.1.0 with the build number 0 does not actually correspond to any particular commit. This is probably undesirable given that coffee and associated packages tend to be under active development.

FabioLuporini:
I think this is good to go. Shall we land it?

kynan:
@FabioLuporini What are thoughts on @gmarkall's concerns? Any plans to tag a COFFEE release?

gmarkall:
Just to follow up on this - merging at the moment would still have the issue that subsequent builds of COFFEE would all have the exact same version and build number. This is a problem because it would mean that running `conda update coffee` would result in conda erroneously determining that coffee is up to date, and users would end up stuck with outdated versions, which in my mind would defeat the purpose of offering conda packages for the components of the Firedrake project (offering an easy way to always install the latest version of the Firedrake project).

I notice that since this PR was made, one release was tagged (`archer-nonaffine`) which I tried to use as the version number for the package, but it didn't succeed because it didn't follow the PEP440 version number convention. There are two ways forward that I can see:

* A release is tagged in this repo with a PEP440 compliant version number (e.g. 0.1.0) - then, I can adjust this recipe to make the version number the tag, and the build number the number of commits since the tag, which is sufficient to make conda's updating mechanism distinguish coffee packages correctly (we do it this way for Numba and other things). This option is strongly desirable because it requires no manual intervention.

* This PR is merged as is, and then every subsequent pull request modifies the recipe in order to increment the version or build number so that each build on Anaconda Cloud (formerly Binstar) has a distinguishing version or build number. (Or alternatively, each subsequent PR does not modify the version or build numbers, and each build is indistinguishable from the last).

Is it possible to tag a commit with something like 0.1.0, or perhaps 0.1.0a0 if you were saving 0.1.0 for a "proper" release?

FabioLuporini:
Thanks for this, @gmarkall 

Forgive my naiveness: since I can't see a clear point in the history that is worth tagging for some special reasons, could we not in principle just tag the latest commit ? From the moment you strongly recommend option 1, that is the one I'm considering here

gmarkall:
Yes, it would be fine to just tag the latest commit on master - as long as this branch is based on the tagged commit (which it will be, because I will rebase this branch) then the versioning mechanism of the conda packaging will work fine.

FabioLuporini:
OK, so I think I created a simple draft release and I granted you write access. I think if you don't have major objections or other things to point out, you can just go ahead in the way you believe to be best (you can then formalize the tag)

gmarkall:
@FabioLuporini Many thanks for this - I formalised the release in order to test the latest commit that I pushed (which used git describe for the version and build number, and required the tag to actually be present). If you are happy to merge this, the I think it is good to go ahead - once it is merged, I will set it up to build on Anaconda cloud so that there's always the latest version available to conda users.

FabioLuporini:
Sounds very good, thank you! Merged.

PR #58, opened 2015-08-05
Log flops

wence-:
Add visitor to estimate the number of effective flops a tree performs

### DIFF
--- a/coffee/visitors/utilities.py
+++ b/coffee/visitors/utilities.py
@@ -6,7 +6,8 @@ import numpy as np
 import itertools
 
 
-__all__ = ["ReplaceSymbols", "CheckUniqueness", "Uniquify", "Evaluate"]
+__all__ = ["ReplaceSymbols", "CheckUniqueness", "Uniquify", "Evaluate",
+           "EstimateFlops"]
 
 
 class ReplaceSymbols(Visitor):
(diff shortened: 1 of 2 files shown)


FabioLuporini:
Makes sense. I think we could also add a simple wrapper function to utils.py so that one doesn't have to use the visitor directly to get the flop count.

FabioLuporini:
Ah, wait. If a loops get vectorized, we should divide by the vector length the impact of *innermost* loops. Or we can return two numbers, the scalar flop count and the vector flop count. To get the vector length, you should use

wence-:
Can you explain what you mean using an example.

wence-:
IOW, I think I do this, whenever I hit a loop, I count the flops in the loop body and then multiply by the number of loop iterations (the size / increment).

So if we have:

```c
for ( i = 0; i < 16; i += 4 ) {
    _mm_add_pd(a, b);
}
```

Then I count each add_pd as 4 flops and multiply by 16 / 4 = 4, giving 16 flops.  Which I think is right.

FabioLuporini:
That doesn't take into account autovectorization of scalar loops. Even if the loop increment is +=1, the loop could (should, is) autovectorized. Depending on whether you're using sse, avx, etc, the loop increments varies. The vector length (and so the autovectorized loop increment) is given by isa['dp_reg'] (double precision per vector register)

wence-:
I don't care about "actual" flops in the compiled code.  I am counting "effective" flops.

wence-:
In particular, I have NO WAY of knowing if the compiler has auto-vectorized the loop, so what am I meant to count?

FabioLuporini:
if you want to estimate the gigaflops per seconds you do, which is what I think you want to do, I don't see how what you call effective flops would be useful.

wence-:
Why not?  I count the number of flops I do, I divide by the time it took.  Lo and behold, "flop/s"

FabioLuporini:
I meant "if vectorization is used"

wence-:
So?  The effective flops are the number of double precision floating point operations I /would have had to do/ to get my answer.  I don't care how they executed.

wence-:
To put it another way.  Doing four double precision floating point multiplies is, in my world, four double precision floating point operations, irrespective of whether it's compiled to four scalar instructions or one avx instruction.  You seem to be arguing for some other count.

FabioLuporini:
Ah, forget it, makes sense. 

I thought we needed the actual number of vector operations to see how many vector operations we were doing so that we could compare it to the corresponding machine peak, but yes, this is not what usually one does

PR #59, opened 2015-08-12
Log flops

wence-:
A few fixes for flop logging.

### DIFF
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -534,6 +534,14 @@ class Statement(Node):
     def __init__(self, children=None, pragma=None):
         super(Statement, self).__init__(children, pragma)
 
+    @property
+    def lvalue(self):
+        return None
+
+    @property
+    def rvalue(self):
+        return None
+
 
 class EmptyStatement(Statement):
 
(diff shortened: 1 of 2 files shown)


FabioLuporini:
@wence : this init property that you added, where do you use it? is it in a branch in PyOP2? what I was referring to with the latest comment is the fact that we should rather use the lvalue and rvalue properties

wence-:
In the `CountFlops` visitor, I ask a for loop for its size (which uses the init property of the iteration variable).  However, in the case that the iteration variable is not declared in the for loop, that currently doesn't exist because an `Assign` object has no init property.

FabioLuporini:
Yes, but "now" (i.e., with that commit) both Decl and Assign have this {l,r}value property. Can you not just use it?

wence-:
Oh, I see, probably.

wence-:
I've now cherry picked that commit and added a few more tests.

FabioLuporini:
OK, if they all pass, let's merge this.

wence-:
Lint fail, have pushed a fix, hopefully this now works.

PR #61, opened 2015-09-17
Rework the optimization engine (Add O4)

FabioLuporini:
This introduces a new optimization engine that aims to automatically apply an "optimal sequence of transformation steps" to arbitrary multilinear forms.

I'm writing a document (paper) to clarify what I mean by "optimal sequence of transformation steps", so I won't waste my (and your) time talking about it here.

I wanna clarify a few practical aspects:

* the new optimisation engine (O4) is essentially a superset of the optimisation mode O2 (O4 "uses" O2). So despite the huge amount of additions brought by this PR, things won't change (i.e. kernels won't be transformed differently) if O2 is actually in place (e.g. current Firedrake). This is on purpose, since O4 may need more time (and more people using it) to become stable. If O4 is switched on, however, all tests (even firedrake ones) at the moment pass, so one could start "beta testing" it.
* many (most!) changes are actually a refactoring or generalisation of ancient code

### DIFF
--- a/coffee/ast_analyzer.py
+++ b/coffee/ast_analyzer.py
@@ -185,3 +185,20 @@ class ExpressionGraph(object):
             elif nx.has_path(self.deps, s.symbol, target_sym.symbol):
                 return True
         return False
+
+    def shares(self, symbols):
+        """Return an iterator of tuples, each tuple being a group of symbols
+        identifiers sharing the same reads."""
+        groups = set()
+        for i in [set(self.reads(s)) for s in symbols]:
+            group = tuple(j for j in symbols if i.intersection(set(self.reads(j))))
+            groups.add(group)
+        return list(groups)
+
+    def readers(self, sym):
+        """Return the list of symbol identifiers that read from ``sym``."""
+        return [i for i, j in self.deps.in_edges(sym)]
+
+    def reads(self, sym):
+        """Return the list of symbol identifiers that ``sym`` reads from."""
+        return [j for i, j in self.deps.out_edges(sym)]
--- a/coffee/ast_analyzer.py
+++ b/coffee/ast_analyzer.py
@@ -121,11 +121,11 @@ class StmtTracker(OrderedDict):
 
     @property
     def all_stmts(self):
-        return set((stmt_info.stmt for stmt_info in self.values()))
+        return set((stmt_info.stmt for stmt_info in self.values() if stmt_info.stmt))
 
     @property
     def all_places(self):
-        return set((stmt_info.place for stmt_info in self.values()))
+        return set((stmt_info.place for stmt_info in self.values() if stmt_info.place))
 
     @property
     def all_loops(self):
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -647,9 +647,13 @@ class Decl(Statement):
         self.qual = qualifiers or []
         self.attr = attributes or []
         self.init = as_symbol(init) if init is not None else EmptyStatement()
+        self._core = self.sym.rank
 
     def operands(self):
         return [self.typ, self.sym, self.init, self.qual, self.attr], {}
+
+    def pad(self, rank):
+        self.sym.rank = rank
 
     @property
     def lvalue(self):
@@ -673,6 +677,15 @@ class Decl(Statement):
           -> ``(20, 10)``)
         """
         return self.sym.rank or (0,)
+
+    @property
+    def core(self):
+        """Return the size of the declaraed variable without including padding."""
+        return self._core
+
+    @core.setter
+    def core(self, val):
+        self._core = val
 
     @property
     def is_const(self):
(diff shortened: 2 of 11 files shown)


FabioLuporini:
@miklos1 I have rebased this, I think, after almost a month, it's time to merge it

FabioLuporini:
Feel free to play with it, check the tests etc. It works on my local branches. I'll wait for a ping from you

coffee/ast_analyzer.py
```diff
@@ -120,8 +120,12 @@ def place(self, sym):
         return self[sym].place if self.get(sym) else None
 
     @property
+    def all_stmts(self):
+        return set((stmt_info.stmt for stmt_info in self.values() if stmt_info.stmt))
```
miklos1: Do we really need double parentheses here?

coffee/base.py
```diff
@@ -675,6 +684,15 @@ def size(self):
         return self.sym.rank or (0,)
 
     @property
```
miklos1: Why is it a property? You could get the same thing by simply reading and writing `self.core`.

coffee/expression.py
```diff
@@ -141,19 +143,29 @@ def is_scalar(self):
     def is_tensor(self):
         return not self.is_scalar
 
+    @property
```
miklos1: ditto

coffee/expression.py
```diff
@@ -39,7 +39,7 @@ class MetaExpr(object):
 
     """Metadata container for a compute-intensive expression."""
 
-    def __init__(self, type, parent, loops_info, domain):
+    def __init__(self, type, parent, loops_info, domain, mode=0):
```
miklos1: What is mode 0? What are the possible values for mode?

coffee/ast_analyzer.py
```diff
@@ -120,8 +120,12 @@ def place(self, sym):
...
+
+    @property
     def all_places(self):
-        return set((stmt_info.place for stmt_info in self.values()))
+        return set((stmt_info.place for stmt_info in self.values() if stmt_info.place))
```
miklos1: ditto

coffee/expression.py
miklos1: Is it the optimisation level?

coffee/loop_scheduler.py
```diff
@@ -215,131 +215,275 @@ def merge(self, root):
...
+
+            self._success = False
+            right = dcopy(node)
+            self._cut(right.children[1], right, 'remainder')
+
+            return left, right
+
+    class CutterSum(Cutter):
+
+        def _cut(self, node, parent, side, topsum=None):
```
miklos1: Why is it not a visitor?

coffee/loop_scheduler.py
```diff
@@ -215,131 +215,275 @@ def merge(self, root):
...
+                return (split, splittable)
+            return ((node, expr_info), ())
+
+    class CutterMatch(Cutter):
+
+        def __init__(self, expr_fissioner):
+            ExpressionFissioner.Cutter.__init__(self, expr_fissioner)
+            self.matched = []
+
+        def _cut(self, node, parent, side, topsum=None):
```
miklos1: Why is it not a visitor?

coffee/loop_scheduler.py
```diff
@@ -215,131 +215,275 @@ def merge(self, root):
...
+            For example, consider the expression a*b + c*d; if the expression is cut
+            into chunks containing only one operand (i.e., self.cut=1), then we have
+            precisely two chunks, /split/ = a*b, /remainder/ = c*d
+
+            If the input expression is a*b + c*d + e*f, and still self.cut=1, then we
+            have two chunks, /split/ = a*b, /remainder/ = c*d + e*f; that is,
+            /remainder/ always contains the subexpression after the fission point
+            """
+            self._success = False
+            left = dcopy(node)
```
miklos1: You should make `_cut` a visitor which reconstructs the tree, and then you can avoid these deep copies.

coffee/ast_analyzer.py
FabioLuporini: no

coffee/expression.py
FabioLuporini: it's in the doc string of the constructor

miklos1: Docstring says "suggested rewrite mode". Can you elaborate? Isn't it the optimisation level?

coffee/loop_scheduler.py
FabioLuporini: Yes, you're right. Unfortunately there is now no time for this now

coffee/expression.py
```diff
@@ -39,7 +39,7 @@ class MetaExpr(object):
...
         """Initialize the MetaExpr.
```
FabioLuporini: Sure. So the rewrite mode is about an expression. The optimisation level includes various things, among which the rewrite mode level.

PR #62, opened 2015-12-07
Fix symbol replacement after loop merge

FabioLuporini:


kynan:
Travis seems to be unhappy.

coffee/base.py
```diff
@@ -445,6 +445,9 @@ def __init__(self, symbol, rank=(), offset=()):
     def operands(self):
         return [self.symbol, self.rank, self.offset], {}
 
+    def reconstruct(self):
+        return Symbol(self.symbol, self.rank, self.offset)
```
wence-: `reconstruct` should take new arguments to reconstruct with.  The default reconstruct method does this.  Why do you need this change (which causes tests to fail)?

FabioLuporini: Ah right, I think I see now how this works. I'll fix this

FabioLuporini:
uops! That was due to leftover code, I have no idea of why it was there. Anyway, purged

wence-:
This is good to go, right?

PR #64, opened 2016-01-15
Fix routine that minimises temporaries

FabioLuporini:


wence-:
You'll need to update the tests to the new CountOccurences interface I think.

wence-:
Oh no, linting!

wence-:
Can you (and/or @alsgregory) formulate the code that creates the problem as a firedrake test so we don't regress?

FabioLuporini:
Oh when I wrote "its interface" I just meant the wrapper in utils.py. The actual interface of the visitor has unchanged

FabioLuporini:
Latest commit should have fixed linting

FabioLuporini:
can we merge this?

wence-:
Yes

wence-:
@alsgregory Can you formulate your problem here as a test case for the firedrake test suite?

PR #67, opened 2016-04-01
TSFC integration

FabioLuporini:
This PR does two things:

- fixes the integration with TSFC and enables expression rewriting
- drastically improves the quality of the codebase

TODO:

- integration of padding + data alignment

Comments welcome 

There are a few pending issues and the need for a little bit of rebasing, but for the rest it should be pretty much ready


dorugeber:
Ugh.  Obvious comment: the commit is +2400/-2300, and there's lots of 'new files added' and 'old files deleted' which I assume is just code shuffling around.  If you actually want someone to read the PR in detail (maybe you don't!), is there a sane way to split this into a trivial "shuffling" PR and a PR where the changes happen?

dorugeber:
(If it would take you an hour, just do it.  If it would take you a week, don't!)

FabioLuporini:
That's true. It's actually just code refactoring. The only real new stuff is in cse.py, plus a few things in hoister.py. The rest is just code move/renamed/refactored.

Thanks

coffee/cse.py
```diff
@@ -0,0 +1,407 @@
+# This file is part of COFFEE
+#
+# COFFEE is Copyright (c) 2014, Imperial College London.
```
miklos1: Wrong year, and possibly wrong copyright holder.

coffee/cse.py
```diff
@@ -0,0 +1,407 @@
...
+
+from plan import verbose
+from base import *
+from utils import *
+from coffee.visitors import EstimateFlops
+from expression import MetaExpr
+
+
+class Temporary():
+
```
miklos1: Docstring?

coffee/cse.py
```diff
@@ -0,0 +1,407 @@
...
+    def __str__(self):
+        return "%s: level=%d, flops/iter=%d, reads=[%s], isread=[%s]" % \
+            (self.symbol, self.level, self.flops,
+             ", ".join([str(i) for i in self.reads]),
+             ", ".join([str(i) for i in self.is_read]))
+
+
+class CSEUnpicker():
+
```
miklos1: Docstring?

coffee/cse.py
```diff
@@ -0,0 +1,407 @@
...
+        if verbose:
+            print BLUE % ("Cost model :: unpicking CSE between levels [%d, %d]:" % bounds),
+            print BLUE % ("cost=%d (cse=%d, outloop=%d, inloop_fact=%d, inloop_cse=%d)" %
+                          (uptolevel_cost, cse_cost, total_outloop_cost,
+                           level_inloop_cost, cse))
+
+        return best
+
+    def unpick(self):
+        fors = visit(self.header, info_items=['fors'])['fors']
```
miklos1: Docstring?

coffee/cse.py
```diff
@@ -0,0 +1,407 @@
...
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT HOLDERS OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+# INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+
```
miklos1: A module docstring would be nice as well.

coffee/expander.py
```diff
@@ -0,0 +1,244 @@
+# This file is part of COFFEE
+#
+# COFFEE is Copyright (c) 2014, Imperial College London.
```
miklos1: copy-paste

coffee/expander.py
```diff
@@ -0,0 +1,244 @@
...
+        return exp
+
+    def invalidate(self, exp):
+        was_hit = False
+        for i, j in self._map.items():
+            if str(j) == str(exp):
+                self._map.pop(i)
+                if self._hits[i] > 0:
+                    was_hit = True
+        return was_hit
```
wence-: AKA:
```python
def invalidate(self, exp):
    for i, j in self._map.items():
        if str(j) = str(exp):
            self._map.pop(i)
            if self._hits[i] > 0:
                return True
     return False
```

coffee/rewriter.py
```diff
@@ -411,593 +446,81 @@ def preevaluate(self):
...
+            # ... define the objective function (min number of factorizations)
+            prob += ilp.lpSum(x[i] for i in nodes_vars)
+
+            # ... solve the problem
+            prob.solve(ilp.GLPK(msg=0))
+
+            # Finally, factorize and hoist (note: the order in which factorizations are carried
+            # out is crucial)
+            nodes = [nodes_vars[n] for n, v in x.items() if v.value() == 1]
+            other_nodes = [nodes_vars[n] for n, v in x.items() if nodes_vars[n] not in nodes]
```
wence-: If I read the docs for pulp correctly, you're meant to pull the answer and the objective out of the solved problem.  Does it also update variables "in place"?

wence-: Also, if the order is crucial is it safe to iterate over a dict and collect things?

coffee/expander.py
```diff
@@ -0,0 +1,244 @@
...
+                return (l_exps + r_exps, self.EXPAND)
+            elif l_type == self.EXPAND and r_type == self.EXPAND and isinstance(node, Sub):
+                return (l_exps + [Neg(r) for r in r_exps], self.EXPAND)
+            else:
+                return ([node], self.GROUP)
+
+        else:
+            raise RuntimeError("Expansion error: unknown node: %s" % str(node))
+
+    def expand(self, should_expand, **kwargs):
```
miklos1: docstring?

coffee/expression.py
```diff
@@ -155,16 +172,11 @@ def mode(self, value):
...
-    ``kwargs``. In particular, ``kwargs`` has the following keys:
-
-        * parent: the block node that embeds the expression.
-        * loops_info: an iterator of 2-tuple ``(loop, loop_parent)`` which
-          substitute analogous information in the new ``MetaExpr``.
-        * domain: the domain of the output tensor evaluated by the expression.
-        * mode: the suggested rewrite mode.
-    """
+    ``kwargs``. ``kwargs`` accepts the following keys: parent, loops_info,
+    linear_dims, mode."""
```
miklos1: Why did you remove the explanation of `parent`, `loops_info` etc. ?

coffee/factorizer.py
```diff
@@ -0,0 +1,233 @@
+# This file is part of COFFEE
+#
+# COFFEE is Copyright (c) 2014, Imperial College London.
```
miklos1: copy-pasta

coffee/factorizer.py
```diff
@@ -0,0 +1,233 @@
...
+
+    @staticmethod
+    def process(symbols, should_factorize, op=None):
+        operands = [s for s in symbols if should_factorize(s)]
+        factors = [s for s in symbols if not should_factorize(s)]
+        return Term(operands, factors, op)
+
+
+class Factorizer():
+
```
miklos1: docstring?

coffee/factorizer.py
```diff
@@ -0,0 +1,233 @@
...
+                    _t.add_factors(factor)
+            factorized = [t.generate_ast for t in factorized.values()]
+            factorized = ast_make_expr(Sum, factorized)
+            parent.children[parent.children.index(node)] = factorized
+            return Term([], [factorized])
+
+        else:
+            return Term([], [node])
+
+    def factorize(self, should_factorize, **kwargs):
```
miklos1: docstring?

coffee/hoister.py
```diff
@@ -0,0 +1,431 @@
+# This file is part of COFFEE
+#
+# COFFEE is Copyright (c) 2014, Imperial College London.
```
miklos1: copy-pasta

miklos1:
This patch removes 22 `dcopy`s and adds 13. I guess that is an improvement.

coffee/rewriter.py
FabioLuporini: Yes. I found it in some tutorials tbh

As for the order: what I actually meant is that all ``nodes`` have to come before all ``other_nodes``

coffee/expander.py
FabioLuporini: just the main expander function (the only one the user is supposed to call, so what it does is basically described in the class' docstring)

coffee/expander.py
FabioLuporini: true :) changed

coffee/expander.py
FabioLuporini: fixed

coffee/cse.py
FabioLuporini: just the main cseunpicker function (the only one the user is supposed to call, so what it does is basically described in the class' docstring)

coffee/cse.py
FabioLuporini: added

coffee/cse.py
FabioLuporini: added

coffee/cse.py
FabioLuporini: fixed the year

coffee/expression.py
FabioLuporini: it was basically the same description as the one you can find in the MetaExpr class' docstring

coffee/factorizer.py
FabioLuporini: fixed

coffee/factorizer.py
FabioLuporini: just the main factorizer function (the only one the user is supposed to call, so what it does is basically described in the class' docstring)

coffee/factorizer.py
FabioLuporini: just the main factorizer function (the only one the user is supposed to call, so what it does is basically described in the class' docstring)

FabioLuporini:
Addressed most of your comments

Please pull

Now the zero-elimination engine is integrated. As a temporary experiment, I've added this transformation to the default optimisation pipeline (i.e., O2) . @wence-  @miklos1  Can you please beta-test this branch at least for functionality ? I'm still running a few experiments for performance. 

Thanks

coffee/factorizer.py
miklos1: meh... how am I supposed to know that? Maybe add some reference at least.

miklos1:
Did you just fix -O4? The hyperelasticity test is unexpectedly passing again.

wence-:
So that xfail was on old pre-tsfc coffee.  I can entirely imagine that many things have changed.  For example, if no optimisations occur then the test xpasses.

miklos1:
With last week's `tsfc-updates` COFFEE, it was still xfailing, but now it passes, and COFFEE claims some flop reduction (though not as much as with -O2).

FabioLuporini:
I've been changing the transformations applied by each optimisation level. 

Now -O4 does what previously -O3 did. Some fixes went it and made it work. 

Also (see last commit in capital letters), at the moment O2 == O3. I've temporarily disabled alignment and padding in {O2,O3} because most users use gcc, and gcc doesn't like auto-vectorising the code, despite all my efforts trying to generate "easily vectorisable source code". Alignment and padding are applied instead at level O4.

I was thinking that one way of deciding whether to apply padding and data alignment is to check whether the underlying compiler is "recent enough" to support the #pragma simd directive (which enforces vectorisation). Most recent versions of gcc have it. We could think of spoofing somehow the compiler version...

coffee/factorizer.py
FabioLuporini: ok done now

coffee/factorizer.py
FabioLuporini: ok doing it now

FabioLuporini:
Added logic for avoiding zero elimination 

Addressed your last comments concerning docstrings

Tests are now running. Maybe it is also now a good time to purge COFFEE from crap ``Par`` nodes :) a git grep seems to show they are never used to construct objects...

.travis.yml
```diff
@@ -10,6 +10,11 @@ language: python
 python:
   - "2.7"
 
+addons:
+  apt:
+    packages:
+    - glpk-utils
```
kynan: Is this going to prevent Travis from running on container infrastructure?

coffee/base.py
```diff
@@ -596,8 +601,13 @@ def rvalue(self, val):
         self.children[1] = val
 
     def gencode(self, not_scope=False):
```
kynan: Double negation? `not_scope=False` == `scope=True`?

coffee/cse.py
```diff
@@ -0,0 +1,419 @@
...
+import operator
+
+from plan import verbose
+from base import *
+from utils import *
+from coffee.visitors import EstimateFlops
+from expression import MetaExpr
+
+
+class Temporary():
```
kynan: New style class?

.travis.yml
FabioLuporini: I honestly have no idea... @miklos1  ?

coffee/base.py
FabioLuporini: yes that's bad... I should maybe change it

coffee/cse.py
FabioLuporini: do you have a preference ?

.travis.yml
miklos1: I mostly added this as documentation, because COFFEE now relies on this, even though its test suite does not actually uses it. I have no idea about Travis.

coffee/cse.py
miklos1: New style class is always better.

miklos1:
Purged `Par` nodes, so this pull request now depends on firedrakeproject/firedrake#746.

coffee/cse.py
```diff
@@ -0,0 +1,419 @@
...
+        cost = 0
+        for level, temporaries in levels.items():
+            cost += sum(t.flops*t.niters for t in temporaries)
+        return cost
+
+    def _cost_fact(self, trace, levels, bounds):
+        # Check parameters
+        bounds = bounds or (min(levels.keys()), max(levels.keys()))
+        assert len(bounds) == 2 and bounds[1] >= bounds[0]
+        assert [i in levels.keys() for i in bounds]
```
kynan: This does't do what you expect it to

wence-: By which he means you wanted `for i in bounds: assert i in levels.keys()`

wence-: Better to write: `assert set(levels.keys()) == set(bounds)` I think.

coffee/cse.py
```diff
@@ -0,0 +1,419 @@
...
+        fact_levels = OrderedDict([(k, v) for k, v in levels.items()
+                                   if k > bounds[0] and k <= bounds[1]])
```
wence-: It's unclear why you would want an ordered dict here given that levels is an unordered dict.

coffee/cse.py
FabioLuporini: OK

coffee/cse.py
FabioLuporini: uooops! Fixing

coffee/cse.py
FabioLuporini: good spot

FabioLuporini:
Addressed latest comments (tests running now)

coffee/expander.py
```diff
@@ -0,0 +1,244 @@
...
+        no new symbols need be introduced. Otherwise (e.g., the H temporary
+        appears in multiple places), create a new symbol."""
+        exp, grp = expansion.left, expansion.right
+
+        # First, check if any of the symbols in /exp/ have been hoisted
+        try:
+            retval = FindInstances.default_retval()
+            exp = [s for s in FindInstances(Symbol).visit(exp, ret=retval)[Symbol]
+                   if s.symbol in self.hoisted and self.should_expand(s)][0]
+        except:
```
kynan: How would the above fail?

FabioLuporini: [0] at the end

miklos1: This is very loose, because `except:` will catch expression thrown from anywhere inside this expression. Can you make it more specific?

wence-: So you're checking if len(...) == 0?

FabioLuporini: yes. OK, I'll make it more specific

coffee/hoister.py
```diff
@@ -0,0 +1,431 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from warnings import warn as warning
+
+from base import *
+from utils import *
+
+
+class Extractor():
```
kynan: New style class?

coffee/factorizer.py
```diff
@@ -0,0 +1,237 @@
...
+                self.factors.remove(f)
+
+    @staticmethod
+    def process(symbols, should_factorize, op=None):
+        operands = [s for s in symbols if should_factorize(s)]
+        factors = [s for s in symbols if not should_factorize(s)]
+        return Term(operands, factors, op)
+
+
+class Factorizer():
```
kynan: New style class?

coffee/factorizer.py
```diff
@@ -0,0 +1,237 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import operator
+
+from base import *
+from utils import *
+
+
+class Term():
```
kynan: New style class?

coffee/hoister.py
```diff
@@ -0,0 +1,431 @@
...
+                # E.g. A[i]*B[j], hoistable in TMP[i,j]
+                return (dep_n, self.EXT)
+        elif info_r == self.EXT:
+            self._try(right, dep_r)
+        elif info_l == self.EXT:
+            self._try(left, dep_l)
+        return (dep_n, self.STOP)
+
+
+class Hoister():
```
kynan: New style class?

coffee/hoister.py
```diff
@@ -0,0 +1,431 @@
...
+
+    # How many times the hoister was invoked
+    _handled = 0
```
kynan: Suppose this functionality could be factored into a common base class?

coffee/rewriter.py
```diff
@@ -31,18 +31,20 @@
...
-from ast_analyzer import ExpressionGraph, StmtTracker
 from coffee.visitors import *
+from hoister import Hoister
+from expander import Expander
+from factorizer import Factorizer
+from cse import CSEUnpicker
 
 
-class ExpressionRewriter(object):
+class ExpressionRewriter():
```
kynan: Why make this an old style class?

coffee/utils.py
```diff
@@ -624,6 +592,184 @@ def to_for(self, itspaces, dims=None, stmts=None):
         return loops
 
 
+###############################################################
+# Utilities for tracking the global impact of transformations #
+###############################################################
+
+
+class StmtTracker(OrderedDict):
```
kynan: I suppose a `StmtTracker` HAS-A `OrderedDict` rather than IS-A `OrderedDict`?

coffee/utils.py
```diff
@@ -624,6 +592,184 @@ def to_for(self, itspaces, dims=None, stmts=None):
...
+
+    whose semantics is, respectively, as follows:
+
+        * The AST node whose ``str(lvalue)`` is used as dictionary key
+        * The AST node of the symbol declaration
+        * The AST node of the closest loop enclosing the statement
+        * The parent of the closest loop
+    """
+
+    class StmtInfo(object):
```
kynan: Use a [`namedtuple`](https://docs.python.org/2/library/collections.html#collections.namedtuple) instead?

requirements.txt
```diff
@@ -1 +1,2 @@
 networkx
+PuLP
```
kynan: Does COFFEE work with the latest PuLP release?

FabioLuporini: The latest is 1.6.1 and it does

coffee/utils.py
FabioLuporini: Ah, I didn't know about it

coffee/utils.py
FabioLuporini: IS-A

wence-: Confusingly, it both *has* an ordereddict and *is* an ordereddict :).

coffee/utils.py
```diff
@@ -624,6 +592,184 @@ def to_for(self, itspaces, dims=None, stmts=None):
...
+                assert(k in self.__class__.INFO)
+                setattr(self, k, v)
+
+    def __init__(self):
+        super(StmtTracker, self).__init__()
+        self.byvalue = OrderedDict()
+
+    def __setitem__(self, key, value):
+        if not isinstance(value, self.StmtInfo):
+            if not isinstance(value, tuple):
```
wence-: `if not isinstance(value, (tuple, self.StmtInfo))`

wence-: Oh, I see why not.

coffee/utils.py
```diff
@@ -624,6 +592,184 @@ def to_for(self, itspaces, dims=None, stmts=None):
...
+                raise RuntimeError("StmtTracker accepts tuple or StmtInfo objects")
+            value = self.StmtInfo(**dict(zip(self.StmtInfo.INFO, value)))
```
wence-: `zip` throws away entries from the longer list.  Do you want to assert that `len(value) == len(self.StmtInfo.INFO`?

coffee/utils.py
FabioLuporini: Ah right I had forgotten about it. It's because I need access by value too at some point

coffee/utils.py
FabioLuporini: adding the check

coffee/expander.py
FabioLuporini: I did better than that, I deleted a whole chunk of code that was only useful in a negligible number of cases

FabioLuporini:
OK, again, addressed your comments.

miklos1:
So it seems that COFFEE only works in addition to the "unroll `IndexSum`s" TSFC optimisation. If that optimisation is disabled (it is on by default), then COFFEE fails. This is not blocker, but it is nice to have it documented.

wence-:
fails, or does nothing?

miklos1:
Fails.

miklos1:
firedrakeproject/firedrake#746 is merged. @FabioLuporini: get this ready for the merge.

Issue #8, opened 2014-12-10
Selecting optimisations a bit odd


wence-:
Imagine I do:

```
parameters['O2'] = True
parameters['O0'] = True
```

And pass these as options to `plan_cpu`, O2 optimisation will be selected.  I wonder if a better way is to have a single "optimisation_level" parameter that defaults to 2.

FabioLuporini:
This is fixed as now one chooses the optimisation level setting 

`parameters['optlevel'] = O0 (O1, O2..)`

I'm closing this

miklos1:
> This is fixed as now one chooses the optimisation level setting
> ```python
parameters['optlevel'] = O0 (O1, O2..)
```

Is this documented anywhere?

FabioLuporini:
coffee.system

wence-:
I think he means, in user-facing docs.  We should have a "misc" section in the firedrake docs about this stuff.

FabioLuporini:
Ah, yes. I'll write something (mostly what's in coffee.system + epsilon)

PR #77, opened 2016-05-31
TSFC + Vectorizer

FabioLuporini:
Make COFFEE capable of applying padding and data alignment to TSFC kernels.

This transformation is compasable with all other transformations. However, it is *not* used by default Firedrake (this is activated at level O3, while Firedrake uses O2). The reason is that despite making most loops easily vectorisable, there are compilers (e.g., GCC ver < 5.0) rejecting auto-vectorisation when loops are too small or due to other non-sense reasons. Padding increases the loop trip counts, so it's (performance-wise) safer to switch it on only when sure that loops are gonna get vectorised.

Note: as I'll document properly and some of you already know, there is a non-trivial interplay between the zero-elimination transformation and padding, because of the introduction of offsets in the kernel (e.g., A[i+3]). This, in some cases, makes data alignment not possible. One should try both O2 and O3 when the performance of a specific kernel is really critical.

Ah, the PR also adds a mechanism to pick the proper citations depending on the set of optimisations applied (didn't bother creating a separate PR). As such, this PR has the (trivial) dependency firedrake #796

### DIFF
--- a/coffee/citations.py
+++ b/coffee/citations.py
@@ -1,0 +1,53 @@
+# This file is part of COFFEE
+#
+# COFFEE is Copyright (c) 2016, Imperial College London.
+# Please see the AUTHORS file in the main source directory for
+# a full list of copyright holders.  All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * The name of Imperial College London or that of other
+#       contributors may not be used to endorse or promote products
+#       derived from this software without specific prior written
+#       permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTERS
+# ''AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT HOLDERS OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+# INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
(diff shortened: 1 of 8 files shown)


coffee/citations.py
```diff
@@ -0,0 +1,54 @@
...
+        nozeros = params.get('dead_ops_elimination', False)
+        align_pad = params.get('align_pad', False)
+        vectorize = params.get('vectorize', False)
+        split = params.get('split', False)
+
+        if rewrite == 1 or align_pad or vectorize or split:
+            Citations().register('Luporini2015')
+        if rewrite > 1 or nozeros:
+            Citations().register('Luporini2016')
+    except:
```
wence-: You should probably only catch `ImportError`.

FabioLuporini: right, fixed

coffee/citations.py
```diff
@@ -0,0 +1,54 @@
+# This file is part of COFFEE
+#
+# COFFEE is Copyright (c) 2014, Imperial College London.
```
miklos1: Wrong year again.

FabioLuporini: Damn...fixed

coffee/plan.py
```diff
@@ -176,6 +178,8 @@ def _generate_cpu_code(self, kernel, **kwargs):
         else:
             params = opts
 
+        update_citations(params)
```
wence-: The only bad thing about this approach is that if we hit the firedrake kernel cache, coffee is never called, and so update_citations(params) is never called.

FabioLuporini: Mm that's true. I could mimic this logic in Firedrake, but Firedrake only "sees" O2 and O3...and that may become a problem as well....?

OK, in Firedrake I could have (I guess in parameters.py):
O1 -> pick Y
O2 -> pick X
O3 -> pick Y 
if something different than O{1,2,3}, then the user is probably in the business of performance optimisation, so we may (i) just rely on the fact that he will find out what the proper citations are or (ii) issue a warning that recalls to clean the cache to get the citations (clean cache => COFFEE will add the citations through the very same mechanism we have now with this PR)

Other options ? Maybe there's a simpler/better way, but this is what I had on top of my head

FabioLuporini: Any idea about this ?

FabioLuporini:
This is now rebased on master

FabioLuporini:
Now depends on PR 817 in Firedrake and PR 492 in PyOP2

FabioLuporini:
This also fixes #8

coffee/system.py
```diff
@@ -85,16 +123,51 @@ def coffee_init(**kwargs):
...
 
     # Allow visits of ASTs that become huge due to transformation. The constant
     # /4000/ was empirically found to be an acceptable value
     sys.setrecursionlimit(4000)
 
 
+def coffee_reconfigure(**kwargs):
+    """Reconfigure the internal state of COFFEE."""
+
+    global options
```
miklos1: Why do you need this?

FabioLuporini: My bad, removed

FabioLuporini:
Prod

PR #78, opened 2016-06-02
Interface clean-up

FabioLuporini:
Straightforward refactoring.

1- Move stuff from plan.py which are not really part of the plan to a different module (system.py)
2- Logging now occurs through a separate module. It's a very basic logging infrastructure, which we might even drop once we centralise the one currently used in PyOP2. But it works for now.

Overall, the quality of the codebase improves

Because of point 1 above, this PR depends on PR 490 in PyOP2

### DIFF
--- a/coffee/logger.py
+++ b/coffee/logger.py
@@ -33,14 +33,8 @@
 
 """The COFFEE logger."""
 
-from collections import OrderedDict
 import logging
 import sys
-try:
-    from mpi4py import MPI
-    rank = MPI.COMM_WORLD.Get_rank()
-except ImportError:
-    rank = 0
 
 
 logger = logging.getLogger('COFFEE')
@@ -68,16 +62,16 @@ RED = '\033[1;37;31m%s\033[0m'
 BLUE = '\033[1;37;34m%s\033[0m'
 GREEN = '\033[1;37;32m%s\033[0m'
 
-COLORS = OrderedDict([
-    (DEBUG, RED),
-    (INFO, GREEN),
-    (COST_MODEL, BLUE),
-    (PERF_OK, GREEN),
-    (PERF_WARN, BLUE),
-    (WARNING, BLUE),
-    (ERROR, RED),
-    (CRITICAL, RED)
-])
+COLORS = {
+    DEBUG: RED,
+    INFO: GREEN,
+    COST_MODEL: BLUE,
+    PERF_OK: GREEN,
+    PERF_WARN: BLUE,
+    WARNING: BLUE,
+    ERROR: RED,
+    CRITICAL: RED
+}
 
 
 def set_log_level(level):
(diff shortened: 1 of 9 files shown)


FabioLuporini:
Now one can disable COFFEE logging in any point of a Firedrake program by simply doing e.g.

from coffee.logger import Logger
Logger.set_min_level()

coffee/logger.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @staticmethod
+    def set_min_level():
+        """Set to minimum log level."""
+        Logger.CURRENT = 'func_warning'
+
+    @staticmethod
+    def set_max_level():
+        """Set to maximum log level."""
+        Logger.CURRENT = 'verbose'
```
wence-: Please just use the Python logging facilities.  See https://docs.python.org/2/library/logging.html

coffee/logger.py
```diff
@@ -0,0 +1,99 @@
+# This file is part of COFFEE
+#
+# COFFEE is Copyright (c) 2016, Imperial College London.
```
miklos1: Good job.

FabioLuporini: Ok

FabioLuporini:
Now use Python's logging module

coffee/logger.py
```diff
@@ -0,0 +1,113 @@
...
+"""The COFFEE logger."""
+
+from collections import OrderedDict
+import logging
+import sys
+try:
+    from mpi4py import MPI
+    rank = MPI.COMM_WORLD.Get_rank()
+except ImportError:
+    rank = 0
```
wence-: Don't do this.  The point of using the python logging is that firedrake can register an appropriate stream handler that does the printing on rank 0 (or not as appropriate).

coffee/logger.py
```diff
@@ -0,0 +1,113 @@
...
+COLORS = OrderedDict([
+    (DEBUG, RED),
+    (INFO, GREEN),
+    (COST_MODEL, BLUE),
+    (PERF_OK, GREEN),
+    (PERF_WARN, BLUE),
+    (WARNING, BLUE),
+    (ERROR, RED),
+    (CRITICAL, RED)
+])
```
wence-: Why does this need to be an ordereddict?

coffee/logger.py
FabioLuporini: OK

coffee/logger.py
FabioLuporini: Uops, leftover from old code

FabioLuporini:
Minor fixes pushed

FabioLuporini:
I think this should just be merged

coffee/logger.py
```diff
@@ -0,0 +1,106 @@
...
+    """Wrapper of the main Python's logging function. Print 'msg % args' with
+    the severity 'level'.
+
+    :arg msg: the message to be printed.
+    :arg level: accepted values are: DEBUG, INFO, COST_MODEL, PERF_OK, PERF_WARN, WARNING,
+        ERROR, CRITICAL
+    """
+    assert level in [DEBUG, INFO, COST_MODEL, PERF_OK, PERF_WARN, WARNING, ERROR, CRITICAL]
+
+    color = COLORS[level] if sys.stdout.isatty() and sys.stderr.isatty() else '%s'
```
wence-: I realise this is merged, but this is terrible.  You get a failure if the log level happens not to match a pre-specified one!

Given that you're only doing this to get the colours, is there a better way?  It appears to be perfectly legit to log with a level that is not one of the predefined ones.

miklos1: > You get a failure if the log level happens not to match a pre-specified one!

We not quite, as the colour does not depend what the log level is set to, but what level `coffee.logger.log` was called with, which is presumably only used from inside COFFEE.

PR #83, opened 2016-07-01
[fix] make sure that we don'T accidentially add the finalizing ; twice

chr-engwer:


chr-engwer:
I ran the test-suite locally and it passes. I guess it is safe to merge.

chr-engwer:
@FabioLuporini should I just merge? Or will you do this?

FabioLuporini:
I'm gonna run it too and then I'll merge it. Does it also pass the Firedrake tests? 

Thanks Christian

chr-engwer:
> I'm gonna run it too and then I'll merge it. Does it also pass the Firedrake tests? 

There was one failing test, but I discussed with others in the office
and we agreed that this is not the fault of my changes, but has some
other reason.

FabioLuporini:
OK, thanks, merging.

PR #84, opened 2016-07-06
Minor fixes

### LINKED ISSUE #65: Cannot import coffee.utils without first importing coffee.visitors
Try the following:

```python
import coffee.utils
```

This fails because of a circular import

If I first do
```python
import coffee.visitors
import coffee.utils
```

Then everything is great.

FabioLuporini:
fixes #65    

@miklos1 this should fix the import error in tsfc


miklos1:
OK

PR #89, opened 2016-08-04
Fix CSE transformer

### LINKED ISSUE #87: derivative of conditional
```
from firedrake import *
# disable_performance_optimisations()

def huber(x, epsilon):
    return conditional(le(x,1),4,x)

m = RectangleMesh(3,3,3,3)
V = VectorFunctionSpace(m, 'CG', 1)

u = Function(V)
v = TestFunction(V)
u.interpolate(Constant((1.0,1.0)))
epsilon = 0.01
E = huber(u[0].dx(0),epsilon)*dx
dE = derivative(E,u,v)
print assemble(dE).vector()[:]
```
Running the code above in Firedrake throws runtime error:
```
RuntimeError: Fission error: found unknown node: (t9 ? 0 : (t11[j] * t6) + (t10[j] * t8))
```
Disabling optimisation fixes the error.

### LINKED ISSUE #88: NaNs with mixed quadrature degree and coffee
So this one is turning out a little tricky to reproduce. But with the following code:

```
from firedrake import *
#parameters['coffee'] = {}
mesh = UnitSquareMesh(1,1)
V = FunctionSpace(mesh, "RT", 2)
P1 = FunctionSpace(mesh, "CG", 1)
v = TestFunction(V)
u = Function(V)
c1 = Function(P1)
c1.assign(0.001)
c2 = Constant(1e-6)
dt = Constant(1.0)
dx2 = dx(degree=5)
F = inner(v,u)*dx - dt*(- c2*inner(v,u)*dx2 - c1*(-u[1]*v[0] + u[0]*v[1])*dx2)
print assemble(derivative(F,u)).M.values.sum()
```

I'm getting NaNs in the assembled matrix (most of the time!) if I leave COFFEE optimisations switched on. I say most of the times, as running the script multiple times I sometimes get different answers (sometimes the correct one), so I expect some sort of initialised value bug? It seems to be also very sensitive to the precise form of the equation.

FabioLuporini:
@miklos1 : finishing to run the firedrake tests now. If they all pass, we should merge this asap

fixes #88, fixes #87

### DIFF
--- a/coffee/cse.py
+++ b/coffee/cse.py
@@ -68,6 +68,11 @@ class Temporary(object):
         return self.symbol.rank if self.symbol else None
 
     @property
+    def is_bilinear(self):
+        linear_loops = [l for l in self.loops if l.dim in self.rank]
+        return len(linear_loops) == 2
+
+    @property
     def symbol(self):
         if isinstance(self.node, Writer):
             return self.node.lvalue
@@ -251,6 +256,8 @@ class CSEUnpicker(object):
         # Code motion
         for t, ew in rewriters.items():
             ew.licm(mode='only_outlinear', lda=lda, global_cse=True)
+            if t.is_bilinear:
+                ew.licm(mode='only_linear')
 
     def _analyze_expr(self, expr, lda):
         finder = FindInstances(Symbol)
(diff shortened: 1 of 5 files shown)


FabioLuporini:
Confirmed: this passes al tests on my machine. @miklos1 can we check and merge this ASAP ? thanks

miklos1:
It would be really nice though if you could make test cases from the reported examples, so that these bugs would not return.

PR #90, opened 2016-08-31
Fix utils makealias

FabioLuporini:


FabioLuporini:
@miklos1  this should be uncontroversial

coffee/base.py
```diff
@@ -435,24 +435,27 @@ class Symbol(Expr):
...
+    def __init__(self, symbol, rank=(), offset=(), operator=None):
         super(Symbol, self).__init__([])
         self.symbol = symbol
         self.rank = rank
         self.offset = offset or tuple([(1, 0) for r in rank])
+        self.operator = operator
 
     def operands(self):
-        return [self.symbol, self.rank, self.offset], {}
+        return [self.symbol, self.rank, self.offset, self.operator], {}
```
wence-: I note that symbols get more and more hair attached to them.  What is it that's going on here?

FabioLuporini: to express things like ``&a`` or ``*a``. operator would then be ``&`` or ``*``

wence-: I think what I mean is that the addition of an additional `operator` to symbols suggests that instead there should be an `Operator` node that attaches this information.  It's already the case that symbols aren't really just symbols but carry either free indices or shape information (depending on context)...

FabioLuporini: I see, and you're right.

I can't easily do this, however, because there are places in which a Symbol is expected, and now an Operator could actually pop up. This clearly denotes a flaw in the design...

What I can do, instead, is extending Symbol with classes such as ``SymbolAddress``, ``SymbolIndirection``, etc which would seamlessly apply the corresponding operators ``&``, ``*`` etc.

Do you want me to proceed this way ?

wence-: I think that would be cleaner.  I realise this it's a lot of work, but we should plausibly consider going round and getting this stuff right.  In particular for some of the hybridisation stuff @thomasgibson is doing, he needs to declare templated C++ types which is painful in the current setup.

miklos1: I think that that ad-hoc inheritance would be even more convoluted than the heavyweight `Symbol` class we currently have.

The right thing to do, of course, is not to be afraid of refactoring.

> there are places in which a Symbol is expected, and now an Operator could actually pop up

Clearly, they can handle more just a symbol if they work with the heavyweight "symbols" that we now have. What are they really expecting? Maybe you can introduce an abstract class for that concept.

Or is it just about skipping certain "symbol modifiers" when checking a subexpression? Then you could just write some utility functions, e.g. `is_modified_symbol`, instead of sticking to `isinstance(..., Symbol)` and lazily adding the kitchen sink to the `Symbol` class.

FabioLuporini: As I said between the lines before, I do agree that some kind of refactoring is what we need here. And probably I'll also have time to work on some of this stuff in the future, as some of my work time might be allocated to coffee-related extensions.

However, the top priority here is to have the huge loop fusion branch be merged *soon*. This fix makes symbols heavier, and this is not good, we all agree on that. Nevertheless, I'd either merge it as it is now or by extending the hierarchy (which is quick to do), because I don't have time *now* to dive into refactoring

FabioLuporini: As an aside note: I'm not saying and I don't think that refactoring that part of the hierarchy would be difficult (operators, shape, etc) , just that it would take a non-negligible (for me, in this period) amount of time

wence-: Sure, I'm happy to just extend the hierarchy for now.

FabioLuporini:
@wence- : done what we discussed earlier

FabioLuporini:
If there are no more objections, I'd merge this.

PR #93, opened 2016-09-22
Handle case when ExpressionGraph.deps has no edges

miklos1:
@FabioLuporini: A single test case on some FInAT branches exercises this code path.


FabioLuporini:
uncontroversial

PR #96, opened 2016-10-17
Add future imports

miklos1:
- Change division to Python 3 divisions from future.
- Enforce through Travis CI that future imports be always added in the future.


wence-:
Looks OK I think.

PR #102, opened 2016-10-27
Extending FunDecl to support template kernel functions

thomasgibson:
This PR is made in anticipation of the SLATE merge coming soon. SLATE's tensor compiler produces template kernel functions of the form: `template <typename Derived> static inline ...` in order to populate the SLATE tensors.

The proposed change leaves the `static inline` check already present in `coffee/plan.py` (originally I removed them), and simply extends `FunDecl` in `base.py` to support an optional `template` parameter for creating template functions.

### DIFF
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -969,25 +969,26 @@ class FunDecl(Statement):
 
     Syntax: ::
 
-        [pred] ret name ([args]) {body};
+        [template] [pred] ret name ([args]) {body};
 
     E.g.: ::
 
         static inline void foo(int a, int b) {return;};"""
 
-    def __init__(self, ret, name, args, body, pred=[], headers=None):
+    def __init__(self, ret, name, args, body, pred=[], headers=None, template=[]):
         super(FunDecl, self).__init__([enforce_block(body)])
         self.pred = pred
         self.ret = ret
         self.name = name
         self.args = args
         self.headers = headers or []
+        self.template = template
 
     def operands(self):
-        return [self.ret, self.name, self.args, self.children[0], self.pred, self.headers], {}
+        return [self.ret, self.name, self.args, self.children[0], self.pred, self.headers, self.template], {}
 
-    def reconstruct(self, ret, name, args, body, pred, headers, **kwargs):
-        return type(self)(ret, name, args, body, pred, headers, **kwargs)
+    def reconstruct(self, ret, name, args, body, pred, headers, template, **kwargs):
+        return type(self)(ret, name, args, body, pred, headers, template, **kwargs)
 
     @property
     def body(self):
@@ -1000,8 +1001,8 @@ class FunDecl(Statement):
     def gencode(self):
         headers = "" if not self.headers else \
                   "\n".join(["#include <%s>" % h for h in self.headers])
-        sign_list = self.pred + [self.ret, self.name,
-                                 wrap(", ".join([arg.gencode(True) for arg in self.args]))]
+        sign_list = self.template + self.pred + [self.ret, self.name,
+                                                 wrap(", ".join([arg.gencode(True) for arg in self.args]))]
         return headers + "\n" + " ".join(sign_list) + \
             "\n{\n%s\n}" % indent(self.children[0].gencode())
 
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -969,20 +969,26 @@ class FunDecl(Statement):
 
     Syntax: ::
 
-        [template] [pred] ret name ([args]) {body};
+        [template]
+        [pred] ret name ([args]) {body};
 
-    E.g.: ::
+    E.g.(Not templated): ::
 
-        static inline void foo(int a, int b) {return;};"""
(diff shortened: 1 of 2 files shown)


coffee/plan.py
```diff
@@ -138,12 +138,24 @@ def plan_cpu(self, opts):
                         # of the expression
                         vect.specialize(*vectorize)
 
+            # First we check if our kernel function is templated
+            templated = False
+            for q in kernel.pred:
+                if q == 'template <typename Derived>':
+                    templated = True
+
```
FabioLuporini: more pythonic:
``templated = any(q == 'template <typename Derived>' for q in kernel.pred)``

FabioLuporini:
A probably cleaner way consists of extending ``FunDecl`` with an additional (optional) argument ``template``, instead of using ``pred``. Then you could properties at will, such as ``is_templated``. 

Also, "Derived" is probably a SLATE keyword, so we should find a way of abstracting from it.

Given these two points, I wonder whether an even better solution is to add a ``class Template`` to ``base.py``, which captures the (minimal) structure of a template, then construct such a node in SLATE and pass it to a FunDecl (as a special argument, not through ``pred``).

Thoughts?

thomasgibson:
@FabioLuporini  I think that's a great idea. I will try to create a class `template` in `base.py` as you suggested as a subclass of `FunDecl` with appropriate methods. I'll update this PR with a commit doing just that. I will update you on this as I work on it.

And yes "Derived" is a keyword that is used in SLATE in compliance with the Eigen C++ library. The `class Eigen::MatrixBase<Derived>` is a base class for all dense matrices, vectors and expressions. SLATE uses this class for populating SLATE tensors, which is why the kernels are templated in this format. It is unlikely to change, as this is the standard template format for the nearly all relevant dense linear algebra functions in Eigen. For functions that take Eigen objects as an argument, I let it take a `MatrixBase` argument as it works for any matrix, vector, or expression objects.

I'll work on creating a class that doesn't particularly care it's a `typename <Derived>` template and instead just recognize it as a general templated function.

thomasgibson:
@FabioLuporini So after playing around with this branch, I found that simply extending `class FunDecl` to support template functions was a small, simple addition (as you first suggested). It's not the `class Template` you mentioned, but it keeps the diff small and it doesn't affect any other bits of COFFEE or Firedrake. What are your thoughts?

miklos1:
I agree with Fabio that we shouldn't make it SLATE/Eigen specific here with the `Derived` template parameter name, and instead should capture a more general template thing. I don't think there's need to invent a new class in this case.

coffee/base.py
```diff
@@ -969,25 +969,26 @@ class FunDecl(Statement):
...
 
-        [pred] ret name ([args]) {body};
+        [template] [pred] ret name ([args]) {body};
 
     E.g.: ::
 
         static inline void foo(int a, int b) {return;};"""
 
-    def __init__(self, ret, name, args, body, pred=[], headers=None):
+    def __init__(self, ret, name, args, body, pred=[], headers=None, template=[]):
```
thomasgibson: Can you elaborate what you mean by "fix `pred`"? How would you rather `pred` be handled?

miklos1: Instead of setting the default value of function arguments a reference to a mutable value (a list in this case), their default value should be `None`, and then you initialise the instance attribute to `pred or []`. This will initialise `self.pred` to a new empty list, rather than the same one.

thomasgibson:
@miklos1 @FabioLuporini I went ahead and pushed the suggested change for `pred` and `template`, as well as added an explicit check for the `template` argument. I didn't squash/remove previous commits yet for the sake of keeping a log of what has changed. Are y'all okay with the body of `def gencode(self)`?

FabioLuporini:
To keep it simple, could you just do ``self.template + self.pred + ...`` ? 

I also wonder whether ``template`` really needs to be a list, since we enforce a maximum length of 1 ?

thomasgibson:
`self.template + self.pred + ...` will only work if `template` is a list, no? That's okay, I can make it a string argument instead and work with that. That would eliminate the check in the `__init__` method. I'll push a change that does just that and you can let me know what you think.

FabioLuporini:
Okay. So now you don't need the ``"" if not self.template else ...`` anymore, right? You can probably just leave ``self.template`` ? 

Then we can merge this :)

thomasgibson:
`"" if not self.template else ...` is actually intentional for cosmetic purposes. While it is the case that the C++ code will compile without the `\n` after the `template` declaration, I was trying to keep it in line with the coding style in C++. However, if we don't care about that, we can just forget it :)

wence-:
I think this is OK.  I hope that coffee handles all the fancy type declarations that will appear later in kernels...

thomasgibson:
Latest push just changes `["" if not self.template else self.template + "\n"]` to `["" if not self.template else (self.template + "\n")]`.

FabioLuporini:
All right. This can be merged as soon as all firedrake tests are confirmed to pass. Then you can go ahead, Thomas. Thanks!

Issue #97, opened 2016-10-21
Invalid loop extent reduction


miklos1:
COFFEE did the following transformation:

``` diff
--- /tmp/pyop2-cache-uid11480/95256e7f5216f51a1d88d64fe2f433bc_p5040.c  2016-10-20 18:15:50.711984816 +0100
+++ /tmp/pyop2-cache-uid11480/16b9d6342d4efc75d2e86893dd6db688_p5237.c  2016-10-20 18:15:25.199611203 +0100
@@ -38,13 +38,23 @@
   for (int  ip_1  = 0; ip_1 < 2; ip_1 += 1)
   {

+    for (int  k0  = 0; k0 < 2; k0 += 1)
+    {
+      
+      for (int  k1  = 0; k1 < 2; k1 += 1)
+      {
+        t8[ip_1][k0][k1] = t2[k0] * t6[ip_1][k1];
+        
+      }
+      
+    }
+    
     for (int  k0  = 0; k0 < 3; k0 += 1)
     {

       for (int  k1  = 0; k1 < 2; k1 += 1)
       {
         t7[ip_1][k0][k1] = t1[k0] * t6[ip_1][k1];
-        t8[ip_1][k0][k1] = t2[k0] * t6[ip_1][k1];

       }  
```

This reduces the extent of the `k0` loop from 3 to 2 for `t8[ip_1][k0][k1] = t2[k0] * t6[ip_1][k1];`. This would be legal, had `t8` been previously initialised to zero, but it wasn't. The loop that was reduced was supposed to initialise it, thus the reduced loop leaving entries uninitialised, and causing all kind of horrible symptoms.

miklos1:
@FabioLuporini: Just wanted to document this, so we won't forget this problem. Not urgent to fix it, since it is not triggered per firedrakeproject/tsfc@1d7ce21.

FabioLuporini:
OK, thanks.

FabioLuporini:
I prefer to keep this open for reference

PR #104, opened 2016-11-25
Hoist reduction loops

### LINKED ISSUE #76: COFFEE results in slower code for this example
I get some blue lines reporting more FLOPs after COFFEE in this example:
https://gist.github.com/colinjcotter/74597eb6e38809ea5e8a585ce203d25c

(requires fix-zero-mixed branch of TSFC).

### LINKED ISSUE #98: Loop merger embarrassed by empty loops
Testing with the `finat` branch of Firedrake, firedrakeproject/tsfc@4589cf2 and `master` FInAT, I get the following failure:

```
$ py.test tests/extrusion/test_helmholtz_scalar.py::test_scalar_convergence[testcase2-3.9-False]
...
../PyOP2/pyop2/host.py:60: in _ast_to_c
    ast_handler.plan_cpu(self._opts)
../COFFEE/coffee/plan.py:120: in plan_cpu
    loop_opt.rewrite(rewrite)
../COFFEE/coffee/optimizer.py:145: in rewrite
    merged_loops = SSALoopMerger(self.expr_graph).merge(self.header)
../COFFEE/coffee/scheduler.py:183: in merge
    merged, l_dims, m_dims = self._merge_loops(parent, l, merging_in)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <coffee.scheduler.SSALoopMerger object at 0x7f8e602b2dd0>
root = <coffee.base.Block object at 0x7f8e5ab0c510>
loop_a = <coffee.base.For object at 0x7f8e5abb6490>
loop_b = <coffee.base.Block object at 0x7f8e5abdb290>

    def _merge_loops(self, root, loop_a, loop_b):
        """Merge the body of ``loop_a`` in ``loop_b`` and eliminate ``loop_a``
            from the tree rooted in ``root``. Return a reference to the block
            containing the merged loop as well as the iteration variables used
            in the respective iteration spaces."""
        # Find the first statement in the perfect loop nest loop_b
        dims_a, dims_b = [], []
>       while isinstance(loop_b.children[0], (Block, For)):
E       IndexError: list index out of range

../COFFEE/coffee/scheduler.py:76: IndexError
```

[Here](https://gist.github.com/miklos1/ac5b794e4301d6a5381933f2e426ce7f) is the unoptimised C code. It seems that the loop merger doesn't expect spurious empty loops.

### LINKED ISSUE #99: Borked loop merging and/or common subexpression elimination
Same test case as in #98, but not generating spurious empty loops per firedrakeproject/tsfc@509cb75, as you can see [here](https://gist.github.com/miklos1/5428ef2dc5bbe8adc8eea97569030dae). This time, however, the test fails because the result is numerically wrong. Looking at the change COFFEE does explains why:

``` diff
--- coffee.c    2016-10-25 12:15:47.165410931 +0100
+++ coffee.c    2016-10-25 12:15:57.533559287 +0100
@@ -57,15 +57,10 @@
   for (int  ip_1  = 0; ip_1 < 4; ip_1 += 1)
   {

-    for (int  i_0  = 0; i_0 < 10; i_0 += 1)
-    {
-      t6[ip_1][i_0] = 0.0;
-      
-    }
-    
     for (int  i_1  = 0; i_1 < 10; i_1 += 1)
     {
-      t7[ip_1][i_1] = 0.0;
+      t6[ip_1][i_1] = 0.0;
+      t7[ip_1][i_1] = t6[ip_1][i_1];

     }

@@ -128,15 +123,10 @@
       double  t35  = 0.0;
       double  t34  = 0.0;

-      for (int  i_0  = 0; i_0 < 10; i_0 += 1)
-      {
-        t34 += t2[ip_0][i_0] * t6[ip_1][i_0];
-        
-      }
-      
       for (int  i_1  = 0; i_1 < 10; i_1 += 1)
       {
-        t35 += t2[ip_0][i_1] * t7[ip_1][i_1];
+        t34 += t2[ip_0][i_1] * t7[ip_1][i_1];
+        t35 += t34;

       }
       double  t36  = ((-1 * t34) + t35);
```

It seems that `t6` and `t7` are incorrectly recognised to be the same (they are not), thus `t2[ip_0][i_0] * t6[ip_1][i_0]` seems to be the same as `t2[ip_0][i_1] * t7[ip_1][i_1]`, and even then the transformation in the lower hunk is just wrong.

### LINKED ISSUE #100: Variable referenced before declaration
Running with the `finat` branch of Firedrake, firedrakeproject/tsfc@c6a55dc, and `master` from FInAT, I'm now getting the classic COFFEE bug:

```
$ py.test tests/extrusion/test_enrichment_1_feec.py::test_feec[horiz_complex0-vert_complex0]
```

See: [9feef36f299985caa0095f4d0a6b8e3f_p7655.{c,err}](https://gist.github.com/miklos1/2cfd715ac9fe8b3b56cf1c959202671f)

### LINKED ISSUE #101: COFFEE does not sum factorise
FInAT `master`, firedrakeproject/tsfc@2fac97b6bd630613dd5f3e5a91ee6b83ab1fe14a (`finat-coffee-patch` branch), running this snippet:

``` python
from ufl import *
from tsfc import compile_form

cell = TensorProductCell(interval, interval)
mesh = Mesh(VectorElement('Q', cell, 1))

f = Coefficient(FunctionSpace(mesh, FiniteElement('Q', cell, 7)))
v = TestFunction(FunctionSpace(mesh, FiniteElement('Q', cell, 7)))

kernel, = compile_form(f*v*dx)

from coffee import O2
from coffee.plan import ASTKernel

knl = ASTKernel(kernel.ast)
knl.plan_cpu(O2)
print knl.gencode()
```

Standard error:

```
Multiple free indices in FlexiblyIndexed: might break COFFEE.
COFFEE finished in 0.0205209 seconds (flops: 19104 -> 19104)
```

[Standard output](https://gist.github.com/miklos1/56f66c74d49354a9359b31b4597b204a).

The multiple free indices complaint is for the coefficient evaluation which is prefactorised by TSFC. For the argument, I reverted the kernel interface on this branch, and fixed the attachment of linear loops pragmas. Yet COFFEE does nothing.

@FabioLuporini: can you have a look at this example, and tell whether:
1. there is a problem with the generated AST and something needs to be generated differently; or
2. COFFEE needs some minor fix in analysis etc. to recognise this; or
3. [Luporini et al. (2016)](https://arxiv.org/abs/1604.05872) does not cover the sum factorisation problem?

Just in case it is unclear, [here](https://gist.github.com/miklos1/7e26788e1709f639b992ef59f8243dc0)'s what the sum factorised version is supposed to look like. If the answer happens to be 3, don't bother wasting too much doing sum factorisation in COFFEE, there is a fairly easy way to do it in TSFC, although that requires argument factorisation. (Fortunately, unlike UFLACS, we will be able to turn that off at least. But it would be nice to have an automatic/optimal logic that gives the best combination of sum factorisation and COFFEE.)

FabioLuporini:
This PR introduces support for hoisting reduction loops (eg, sum-factorisation in TSFC kernels), alongside a number of additional contributions:
* Simplification and generalisation of hoisting algorithms, which now work with loop nests of arbitrary depth
* Generalisation of sharing-graph-based expression rewriting, which now works with n-linear forms (and not just for n <= 2)
* Generalisation of the CSE's cost model, which now handles loop nests of arbitrary depth (this was necessary given the tensor product structure of basis functions)

Also:
* Some obsolete code has been dropped.
* Some ancient code has been refactored.
* Various components and utility functions have been improved/enhanced/fixed.

fixes #76, fixes #98, closes #99, fixes #100, resolves #101


coffee/base.py
```diff
@@ -466,6 +466,14 @@ def is_const(self):
         return not self.rank or all(is_const_dim(r) for r in self.rank)
 
     @property
+    def is_number(self):
```
miklos1: We could have a specific number type if you wish...

coffee/base.py
```diff
@@ -1209,6 +1222,26 @@ def gencode(self, not_scope=False):
         return self.children[0].gencode()
 
 
+class Rank(tuple):
+
+    def __contains__(self, val):
```
miklos1: Are you happy with the usage that sticks an COFFEE expression into `rank`?

coffee/visitors/utilities.py
```diff
@@ -284,7 +283,9 @@ def default_retval(cls):
     """
 
     def __init__(self, symbols):
+        from coffee.utils import flatten
         self.symbols = symbols
+        self.flatten = flatten
```
miklos1: This is odd, why do you assign a static function onto `self`?

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
 ignore =
-    E501,F403,F405,E226,E265,E731,E402,E266,
+    E501,F403,F405,E226,E265,E731,E402,E266,F999,
```
miklos1: What fails without F999? On my machine flake8 is still happy without this option.

coffee/rewriter.py
```diff
@@ -442,40 +543,55 @@ def preevaluate(self):
...
+        self.expand(mode='all')
+
+        # Make sure that potential reductions are not hidden away
+        lda = loops_analysis(self.header, value='dim')
+        self.reassociate(lambda i: (not lda[i]) + lda[i].issubset(set(other_dims)))
+
+        # Construct the sharing graph
+        nodes, edges = [], []
+        for i in summands(self.stmt.rvalue):
+            symbols = zip(*explore_operator(i))[0] if not isinstance(i, Symbol) else [i]
```
miklos1: If you merge with master, you get Python 3 style `zip` from `six.moves`, which means this doesn't work. You need to write `list(zip(...))[0]` instead of `zip(...)[0]`.

setup.cfg
FabioLuporini: Maybe because you have a different global setup? I get hundreds of warnings without F999

coffee/rewriter.py
FabioLuporini: good catch, will fix

coffee/visitors/utilities.py
FabioLuporini: work around to circular imports, as I'm importing ``utils`` and ``utils`` is importing some visitors. Not sure where I should move flatten ? the other option is to re-import flatten in the member methods that use it

coffee/base.py
FabioLuporini: I have a list of changes for COFFEE in mind including a restructuring of ``base.py``, as well as the addition of new types (eg ``Number``) and class attributes (eg ``is_Symbol``, ``is_Incr``). I'll do this in another PR

coffee/base.py
miklos1: Fine, I do not mind having the interface cleanup (which will require corresponding changes in other components) in a separate PR after this one is merged.

setup.cfg
miklos1: I have a global setup? Are you sure your flake8 is up to date?

FabioLuporini: ``
flake8 --version
3.2.1 (pyflakes: 1.3.0, mccabe: 0.5.2, pycodestyle: 2.2.0) CPython 2.7.12 on Linux
``

I think with flake8 you can also have a global config file wherein you state all of the options that you want to be ignored in *any* project -- but maybe that's not your case

coffee/rewriter.py
FabioLuporini: so, let me understand. I'm using ``zip`` that way, in a number of places. I guess I should then fix them all

coffee/visitors/utilities.py
miklos1: > Not sure where I should move flatten ?

`coffee.base` might work I think, although it feels a bit arbitrary. Or you could drop `flatten` altogether and use chain-star instead:

`flatten(xs)` => `list(chain(*xs))` or `list(chain.from_iterable(xs))`.

If you directly use this in a `for`, then you don't even need the `list`. Of course, you will need `from itertools import chain`.

> the other option is to re-import flatten in the member methods that use it

I think that would express the intention (of breaking circular imports) more clearly than assigning to `self`.

coffee/rewriter.py
miklos1: I think I fixed most cases on master. This usage of zip, of course, works in Python 2 if you don't have `from six.moves import zip`. So I suggest to fix these in the additions of this patch, and I also suggest to add the corresponding `six.moves` import (filter, map, range, zip -- if used in the file).

FabioLuporini: Done

coffee/visitors/utilities.py
FabioLuporini: I'm now importing it directly in the relevant member, that is indeed better.

I didn't know about ``chain``; will probably use it all over the place in a next PR

coffee/utils.py
```diff
@@ -451,6 +480,33 @@ def find_expression(node, type=None, dims=None, in_syms=None, out_syms=None):
...
+    ========
+
+    a + b --> [a, b]
+    a*b*c --> [a*b*c]
+    a*b*c + c*d --> [a*b*c, c*d]
+    (a+b)*c + d --> [(a+b)*c, d]
+    foo(a) --> []
+    """
+
+    handle = zip(*explore_operator(node))
```
miklos1: Suggested fix: cast iterable to list. The iterable is always "true" in a condition, whether it will yield any elements or not.

FabioLuporini: Done

miklos1:
When testing against master branches, I get failures in Firedrake tests/regression/test_conditional.py

miklos1:
The conditional problem seems to be fixed, for example, with this patch:
```diff
diff --git a/coffee/expander.py b/coffee/expander.py
index 9d41b40..9ece54c 100644
--- a/coffee/expander.py
+++ b/coffee/expander.py
@@ -73,7 +73,7 @@ class Expander(object):
             return ([node], self.EXPAND) if self.should_expand(node) \
                 else ([node], self.GROUP)
 
-        elif isinstance(node, (Div, FunCall)):
+        elif isinstance(node, (Div, Ternary, FunCall)):
             # Try to expand /within/ the children, but then return saying "I'm not
             # expandable any further"
             for n in node.children:
diff --git a/coffee/hoister.py b/coffee/hoister.py
index 52147d4..5da6d9b 100644
--- a/coffee/hoister.py
+++ b/coffee/hoister.py
@@ -83,7 +83,7 @@ class Extractor(object):
 
         elif isinstance(node, (FunCall, Ternary)):
             arg_deps = [self._visit(n) for n in node.children]
-            dep = tuple(set(flatten([dep for dep, _ in arg_deps])))
+            dep = set(flatten([dep for dep, _ in arg_deps]))
             info = self.EXT if all(i == self.EXT for _, i in arg_deps) else self.STOP
             return (dep, info)
 
diff --git a/coffee/rewriter.py b/coffee/rewriter.py
index 289b150..b20e8fa 100644
--- a/coffee/rewriter.py
+++ b/coffee/rewriter.py
@@ -409,7 +409,7 @@ class ExpressionRewriter(object):
             if isinstance(node, (Symbol, Div)):
                 return
 
-            elif isinstance(node, (Sum, Sub, FunCall)):
+            elif isinstance(node, (Sum, Sub, FunCall, Ternary)):
                 for n in node.children:
                     _reassociate(n, node)
 
```

miklos1:
There is another issue against master branches:
```c
  ...

  A[0][0] = A00[0][0];
  A[0][1] = A01[0][0];
  A[0][2] = A00[0][1];
  A[0][3] = A01[0][1];
  A[1][0] = A10[0][0];
  A[1][1] = A11[0][0];
  A[1][2] = A10[0][1];
  A[1][3] = A11[0][1];
  A[2][0] = A00[1][0];
  A[2][1] = A01[1][0];
  A[2][2] = A00[1][1];
  A[2][3] = A01[1][1];
  A[3][0] = A10[1][0];
  A[3][1] = A11[1][0];
  A[3][2] = A10[1][1];
  A[3][3] = A11[1][1];
  
  for (int  j  = 0; j < 2; j += 1)
  {
    for (int  k  = 0; k < 2; k += 1)
    {
      #pragma coffee expression
      A00[j][k] += (t0[j] * ct5[k]) + (t1[j] * ct6[k]);   
    }
  }
  
  for (int  j  = 0; j < 2; j += 1)
  {
    for (int  k  = 0; k < 2; k += 1)
    {
      #pragma coffee expression
      A11[j][k] += (t0[j] * ct12[k]) + (t1[j] * ct13[k]);
    }
  }
```

Obviously, the `A[j][k] = ...` copying should remain at the end of the function. This does not happen with FInAT, since that doesn't need this rearrangement hack for interior facet integrals with vector arguments.

FabioLuporini:
I hadn't tested with firedrake master yet.

The patch looks fine (the ``tuple(set(...`` --> ``set(...`` is a good catch)

Tell me how to reproduce the other bug, which should be an easy fix

miklos1:
> Tell me how to reproduce the other bug, which should be an easy fix

firedrake tests/extrusion/test_interior_facets_extr.py (use master branches)

miklos1:
Does not actually fix #97, but I don't think it matters any more. We may still close that as wontfix because of firedrakeproject/tsfc@1d7ce21.

FabioLuporini:
I'm also happy to land this. Is that fine for you if I just hit merge, or do you prefer to go in synch with some other PRs?

miklos1:
I'm fine with just merging, I was just wondering if @wence- has anything to say.

PR #105, opened 2016-11-30
Python 3 compatibility fixes

miklos1:
`.keys()`, `.items()`, `.values()` of `dict` return iterables rather than lists in Python 3.

### DIFF
--- a/coffee/cse.py
+++ b/coffee/cse.py
@@ -32,6 +32,7 @@
 # OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from __future__ import absolute_import, print_function, division
+from six import iterkeys, iteritems, itervalues
 from six.moves import zip
 
 import operator
@@ -99,7 +100,7 @@ class Temporary(object):
 
     @property
     def linear_reads(self):
-        return self.linear_reads_costs.keys() if self.linear_reads_costs else []
+        return list(iterkeys(self.linear_reads_costs)) if self.linear_reads_costs else []
 
     @property
     def loops(self):
@@ -205,11 +206,11 @@ class CSEUnpicker(object):
 
     @property
     def type(self):
-        return self.exprs.values()[0].type
+        return list(itervalues(self.exprs))[0].type
 
     @property
     def linear_dims(self):
-        return self.exprs.values()[0].linear_dims
+        return list(itervalues(self.exprs))[0].linear_dims
 
     def _push_temporaries(self, temporaries, trace, global_trace, ra, decls):
 
@@ -261,14 +262,14 @@ class CSEUnpicker(object):
         # Transform the AST (note: node replacement must happen in the order
         # in which the temporaries have been encountered)
         modified_temporaries = sorted(modified_temporaries.values(),
-                                      key=lambda t: global_trace.keys().index(t.urepr))
+                                      key=lambda t: list(iterkeys(global_trace)).index(t.urepr))
         for t in modified_temporaries:
             ast_replace(t.node, to_replace, copy=True)
         replaced = [t.urepr for t in to_replace.keys()]
 
         # Update the temporaries
         for t in modified_temporaries:
-            for r, c in t.linear_reads_costs.items():
+            for r, c in list(iteritems(t.linear_reads_costs)):
                 if r.urepr in replaced:
                     t.linear_reads_costs.pop(r)
                     r_linear_reads_costs = global_trace[r.urepr].linear_reads_costs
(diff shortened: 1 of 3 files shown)


coffee/cse.py
```diff
@@ -206,11 +207,11 @@ def __init__(self, exprs, header, hoisted, decls):
 
     @property
     def type(self):
-        return self.exprs.values()[0].type
+        return list(itervalues(self.exprs))[0].type
```
wence-: This implicitly (before it was also the case) assumes that self.exprs has one entry, no?

coffee/cse.py
```diff
@@ -411,7 +412,7 @@ def _cost_fact(self, trace, levels, lda, bounds):
                     else:
                         handle = [read]
                     linear_reads.extend(handle)
-                factors = {as_urepr(i): i for i in linear_reads}.values()
+                factors = list(itervalues({as_urepr(i): i for i in linear_reads}))
```
wence-: AKA `linear_reads`.

wence-: Oh, no, it's eliminating duplicates, but in a non-deterministic order.

coffee/cse.py
FabioLuporini: yes otherwise you wouldn't reach that point

coffee/cse.py
FabioLuporini: you don't need determinism there because the info is just used by the cost model to eventually produce a number

coffee/scheduler.py
```diff
@@ -835,7 +836,7 @@ def reschedule(self, root):
             # Split the main expressions to maximize the impact of the rescheduling (this
             # helps if different summands have zero-valued blocks at different offsets)
             elf = ExpressionFissioner(cut=1, loops='none')
-            for stmt, expr_info in self.exprs.items():
+            for stmt, expr_info in list(iteritems(self.exprs)):
                 if expr_info.is_scalar:
                     continue
                 self.exprs.pop(stmt)
```
FabioLuporini: it's fun because the more I look at it, and the less I understand. Bho. Sometimes I do things and I simply don't know why I do them that way.

blechta:
With this branch and Py3

```
from ufl import *
from tsfc import compile_form
from coffee.plan import ASTKernel

element = FiniteElement("P", interval, 1)
u, v = TrialFunction(element), TestFunction(element)
a = inner(grad(u), grad(v))*dx
c = compile_form(a)
a = ASTKernel(c[0].ast)
a.plan_cpu(dict(optlevel='O2'))
print(a.gencode())
from ufl import *
from tsfc import compile_form
from coffee.plan import ASTKernel

element = FiniteElement("P", interval, 1)
u, v = TrialFunction(element), TestFunction(element)
a = inner(grad(u), grad(v))*dx
c = compile_form(a)
a = ASTKernel(c[0].ast)
a.plan_cpu(dict(optlevel='O2'))
print(a.gencode())
```

fails with

```
  File "/home/jan/dev/fenics-master/src/COFFEE/coffee/plan.py", line 119, in plan_cpu
    loop_opt.rewrite(rewrite)
  File "/home/jan/dev/fenics-master/src/COFFEE/coffee/optimizer.py", line 104, in rewrite
    self._unpick_cse()
  File "/home/jan/dev/fenics-master/src/COFFEE/coffee/optimizer.py", line 174, in _unpick_cse
    cse_unpicker.unpick()
  File "/home/jan/dev/fenics-master/src/COFFEE/coffee/cse.py", line 508, in unpick
    for transformed_loop, nest in reversed(nests.items()):
TypeError: argument to reversed() must be a sequence
```

miklos1:
@blechta: what am I doing wrong? your example just works for me on this branch...

blechta:
Ok, just get it working in a container with Python 3.5.2. Is it possible that it does not work on Python 3.4.3?

blechta:
This is the code which works with Py3.5 but not with Py3.4:

```
from collections import OrderedDict
d = OrderedDict(foo=1, bar=2)
reversed(d.items())
```

Would you accept a reasoble workaround if I provide one?

miklos1:
Sure. You could just replace `foo.items()` with `list(iteritems(foo))` where `from six import iteritems` and it should work then.

wence-:
Or you could replace:

```python
for transformed_loop, nest in reversed(nests.items()):
            for loop, parent in nest:
                if loop == transformed_loop and not loop.body:
                    parent.children.remove(loop)

```

With
```python
while nests:
    transformed_loop, nest = nests.popitem()
    ...
```

Of course, you'd like to write:
```
while k, v = x.popitem():
   ...
```

But Guido doesn't believe in any kind of destructuring-bind or statement-is-expression semantics.

blechta:
Can you cherry-pick blechta/COFFEE@e0e339ca4ea33bbcba7eae345241b460cd6a7cc5?

Here's the diff of Py2 and Py3 output in FFC regression suite [error.log.txt](https://github.com/coneoproject/COFFEE/files/627901/error.log.txt) with this commit.

miklos1:
@blechta: Done.

coffee/cse.py
FabioLuporini: it's an ``OrderedDict``, yes

PR #112, opened 2016-12-14
Resolving SLATE conflicts in type checking

thomasgibson:
This is a minor PR that attempts to make COFFEE compatible with SLATE. The problem lies in `def check_type` in `coffee/utils.py`. The type of `lhs_decl` will always be `"Eigen::MatrixBase<Derived>"` in the SLATE case. Rather than hard-coding the type check for the eigen matrix, `kernel.template` is examined instead. This is an indirect way to address the SLATE case, as kernels will always be templated.

### DIFF
--- a/coffee/plan.py
+++ b/coffee/plan.py
@@ -93,7 +93,11 @@ class ASTKernel(object):
                 parent, nest = expr_info
                 if not nest:
                     continue
-                metaexpr = MetaExpr(check_type(stmt, info['decls']), parent, nest)
+                if kernel.template:
+                    typ = "double"
+                else:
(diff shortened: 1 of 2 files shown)


coffee/plan.py
```diff
@@ -93,7 +93,11 @@ def plan_cpu(self, opts):
                 parent, nest = expr_info
                 if not nest:
                     continue
-                metaexpr = MetaExpr(check_type(stmt, info['decls']), parent, nest)
+                if kernel.template:
+                    typ = "double"
```
wence-: This works "for now", but note going forward this should be parameterised by the scalar type of the output argument of the kernel I think.

PR #116, opened 2017-03-08
vectorizer: Fix introduction of aligned(...)

FabioLuporini:
*** Maybe not yet ready for merge (read below) ***

Alignment qualifiers now appear in the code.

Still need to run the test suite. 

Let's however start talking about the following "problem":

The vectorizer doesn't like the following type of notation ``A[(j0 * 6) + j1][(k0 * 6) + k1] = ...``
In essence, to maximize the effectiveness of the vector code, it is fundamental that the base addresses of all LOAD/STORE instructions are aligned to multiples of the vector length. COFFEE attempts to achieve this through several transformations, including:
- padding
- data alignment
- loop bounds rounding

Padding and data alignment now look OK (modulo happy Travis), but the innermost loop is sometimes not rounded when instead it could actually be rounded (eg, by rounding the endpoint from 6 to 8 in case AVX is used). While most loops are properly rounded, the loop enclosing ``A[(j0 * 6) + j1][(k0 * 6) + k1] = ...`` is not: this is because COFFEE cannot determine whether the rounding pass will break the computation, as it doesn't know how to interpret array accesses when these are multidimensional functions.

The good news is that the code is anyway vectorizable -- it might just run a little slower than ideal as a remainder loop might be inserted, and executed with SSE instructions instead of AVX (or higher) instructions

### DIFF
--- a/coffee/plan.py
+++ b/coffee/plan.py
@@ -38,7 +38,8 @@ import coffee
 from .base import *
 from .utils import *
 from .optimizer import CPULoopOptimizer, GPULoopOptimizer
-from .vectorizer import LoopVectorizer, VectStrategy
+from .vect2 import LoopVectorizer, VectStrategy
+# from .vectorizer import LoopVectorizer, VectStrategy
 from .expression import MetaExpr
 from .logger import log, warn, PERF_OK, PERF_WARN
 from coffee.visitors import Find, EstimateFlops
--- a/coffee/vect2.py
+++ b/coffee/vect2.py
@@ -1,0 +1,368 @@
+# This file is part of COFFEE
+#
+# COFFEE is Copyright (c) 2014, Imperial College London.
+# Please see the AUTHORS file in the main source directory for
+# a full list of copyright holders.  All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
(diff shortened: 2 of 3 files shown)


wence-:
I really hope this loop doesn't get executed with SSE, since if the rest is done with avx (for whatever reason), we'll get horrbile performance: switching from avx to sse incurs multiple 10s of cycles delay.

miklos1:
> The vectorizer doesn't like the following type of notation `A[(j0 * 6) + j1][(k0 * 6) + k1] = ...`

miklos1:
> The vectorizer doesn't like the following type of notation `A[(j0 * 6) + j1][(k0 * 6) + k1] = ...`

It's just an inline reshape. What would you do if it were `A[j0][j1][k0][k1] = ...` where 0 <= `j0` < 6 and 0 <= `k0` < 6?

FabioLuporini:
> What would you do if it were A[j0][j1][k0][k1] = ... where 0 <= j0 < 6 and 0 <= k0 < 6?

you just care about the innermost dimension (k1)

> I really hope this loop doesn't get executed with SSE, since if the rest is done with avx (for whatever reason), we'll get horrbile performance: switching from avx to sse incurs multiple 10s of cycles delay.

your hope is well posed with the Intel compiler (I actually now remember that I had read this in the past): https://software.intel.com/en-us/articles/avoiding-avx-sse-transition-penalties
Read section 3.1. Not sure what happens with GCC. I guess pretty much the same thing.

miklos1:
This cannot be merged yet as it breaks Firedrake.

wence-:
This will soon become obsolete with new loopy codegen.

PR #117, opened 2017-03-28
New optimisation level 'Ov'

tj-sun:
which just does vectorisation and no loop motion

### DIFF
--- a/coffee/__init__.py
+++ b/coffee/__init__.py
@@ -41,7 +41,7 @@ from coffee.system import architecture, compiler, isa
 from coffee.system import set_architecture, set_compiler, set_isa
 
 
-__all__ = ['options', 'initialized', 'O0', 'O1', 'O2', 'O3']
+__all__ = ['options', 'initialized', 'O0', 'O1', 'O2', 'O3', 'Ov']
 
 
 class Options(dict):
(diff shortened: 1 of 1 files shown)


miklos1:
If we are following the tradition of C compilers for naming optimisation levels, then I think we should follow it faithfully and use lowercase letters for the level "number", i.e. `Ov` instead of `OV`, similarly to `-Os` or `-Og` which are common to C compilers.

miklos1:
Commit message has a typo and its first line is too long.

coffee/__init__.py
```diff
@@ -174,6 +174,7 @@ def set_opt_level(optlevel):
 O1 = OptimizationLevel('O1', rewrite=1)
 O2 = OptimizationLevel('O2', rewrite=2, dead_ops_elimination=True)
 O3 = OptimizationLevel('O3', align_pad=True, **O2)
+OV = OptimizationLevel('Ov', align_pad=True)
```
miklos1: Another big `OV`

tj-sun: I'm surprised how this didn't give an error.. but amended.

miklos1: There is still a big `OV` in the diff...

miklos1:
Can you squash these commits and fix the commit message?

Plus, you have at least one more big OV...

PR #125, opened 2019-04-16
Minimal

wence-:
This strips out all the code apart from the AST and `EstimateFlops`. Moving forward, the goal is that the final bits of kernel optimisation should eventually be ported into TSFC, and the kernel representation will instead of COFFEE-ast be loopy.

If we want to leave `master` full-fat, we can keep this as a branch and use that in firedrake.


miklos1:
Any reason this has not been merged yet?

wence-:
No reason I think. I didn't get roundtoits

